PR #41, opened 2019-12-02
CRAN submission

DominiqueMakowski:
submitted


DominiqueMakowski:
```Please shorten the title to a maximum of 65 characters.
Acronyms can be used on their own in the title as long as they are
explained in the description field.

Please add \value to .Rd files that are not data files and explain the
functions results in the documentation.
f.i.: smoothing.Rd
If a function does not return a value, please document that too, e.g.
\value{None}.
```

DominiqueMakowski:
CRAN reviewers want us to rename the package................. because apparently it's too general..... I answered that it's in line with other packages (performance, parameters) and that is related to the master function (`estimate_*`) but no...

I could either try again and hope that each reaches another reviewer (every time it's a different one that makes different comments) or think about renaming it... 

what are your thoughts?

strengejacke:
hm... a pity, but then I'd vote for renaming.

mattansb:
Booo.... :(((

mattansb:
Maybe `estimates` (with the "s"), as it could be argued we don't estimate anything *per se*, but it provides estimates that were estimated by other means / models? (Stretching this too thin?)

DominiqueMakowski:
I was thinking about something like `modelbased`, so it would be `modelbased::estimate_means()` to underline the fact that this package is made for estimating stuff from models?

strengejacke:
sounds good!

strengejacke:
do you preserve model-class names? (like `"see_estimate_contrasts"`)

DominiqueMakowski:
Yes I'm thinking about changing nothing else than the package name (all functions and classes would remain the same)

PR #56, opened 2020-04-29
Dev

DominiqueMakowski:
back to work


DominiqueMakowski:
do we have *any* travis that passes? 😅

strengejacke:
> do we have _any_ travis that passes? 😅

No ;-)
I think most Linux-builds are not working.

DominiqueMakowski:
I tried adding the GH-CI by adding:
```r
usethis::use_github_action_check_release()
usethis::use_github_action_check_full()
```
let's see

DominiqueMakowski:
failed for 3.3 and 3.4 because:
```
* checking package dependencies ... ERROR
##[error]Package required but not available: ‘emmeans’

Packages suggested but not available: ‘merTools’ ‘rstanarm’ ‘brms’

The suggested packages are required for a complete check.
Checking can be attempted without them by setting the environment
variable _R_CHECK_FORCE_SUGGESTS_ to a false value.
```

PR #64, opened 2020-05-23
dev

DominiqueMakowski:


DominiqueMakowski:
@mattansb we have a problem, it seems like `bayestestR::describe_posterior()` doesn't take into account the transformation 😱 

``` r
      model <- rstanarm::stan_glm(vs ~ mpg, data=mtcars, family="binomial", refresh=0)
      means <- emmeans::emmeans(model, "mpg", type = "response")
      means

#>   mpg  prob lower.HPD upper.HPD
#>  20.1 0.448     0.224     0.688
#> 
#> Point estimate displayed: median 
#> Results are back-transformed from the logit scale 
#> HPD interval probability: 0.95


      bayestestR::describe_posterior(means)

#> # Description of Posterior Distributions
#> 
#> Parameter | Median |          89% CI |    pd |        89% ROPE | % in ROPE
#> --------------------------------------------------------------------------
#> 20.090625 | -0.207 | [-0.999, 0.611] | 0.668 | [-0.100, 0.100] |    16.596


      summary(data.frame(emmeans::as.mcmc.emmGrid(means, names = FALSE, sep.chains = FALSE)))
#>    X20.090625     
#>  Min.   :-2.1361  
#>  1st Qu.:-0.5289  
#>  Median :-0.2069  
#>  Mean   :-0.2096  
#>  3rd Qu.: 0.1142  
#>  Max.   : 1.6399
```


<sup>Created on 2020-05-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

mattansb:
Yes....... As you can see, `as.mcmc.emmGrid` returns everything on the linear scales.

There is a solution for this:
```R
bayestestR::describe_posterior(emmeans::regrid(means))
#> # Description of Posterior Distributions
#> 
#> Parameter | Median | CI | CI_low | CI_high | pd | ROPE_CI | ROPE_low | ROPE_high | ROPE_Percentage
#> --------------------------------------------------------------------------------------------------
#> 20.090625 |  0.448 | 89 |  0.258 |   0.634 |  1 |      89 |   -0.100 |     0.100 |               0

```
`regrid`'s default behavior is to transform the grid to the response (regardless of the call to `emmeans`):
```R
means <- emmeans::emmeans(model, "mpg", type = "link")
means
#>   mpg emmean lower.HPD upper.HPD
#>  20.1  -0.21     -1.19     0.758
#> 
#> Point estimate displayed: median 
#> Results are given on the logit (not the response) scale. 
#> HPD interval probability: 0.95 
 
bayestestR::describe_posterior(emmeans::regrid(means))
#> # Description of Posterior Distributions
#> 
#> Parameter | Median | CI | CI_low | CI_high | pd | ROPE_CI | ROPE_low | ROPE_high | ROPE_Percentage
#> --------------------------------------------------------------------------------------------------
#> 20.090625 |  0.448 | 89 |  0.258 |   0.634 |  1 |      89 |   -0.100 |     0.100 |               0
```
Which is why I didn't implement this in `bayestestR`... Do you think I should?

DominiqueMakowski:
I fixed in bayestestR:

``` r
model <- rstanarm::stan_glm(vs ~ mpg, data=mtcars, family="binomial", refresh=0)

means <- emmeans::emmeans(model, "mpg", type = "response")
means
#>   mpg  prob lower.HPD upper.HPD
#>  20.1 0.452     0.227     0.689
#> 
#> Point estimate displayed: median 
#> Results are back-transformed from the logit scale 
#> HPD interval probability: 0.95
bayestestR::describe_posterior(means)
#> # Description of Posterior Distributions
#> 
#> Parameter | Median |         89% CI | pd |        89% ROPE | % in ROPE
#> ----------------------------------------------------------------------
#> 20.090625 |  0.452 | [0.259, 0.634] |  1 | [-0.100, 0.100] |         0
```

<sup>Created on 2020-05-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

mattansb:
@DominiqueMakowski I am unfamiliar with the inner workings of `modelbased`, but I suggest setting up some unit tests for Bayesian estimates / contrasts (especially in glms) asking for BFs - it is not possible to compute BFs on transformed `emmGrid` objects (via `regrid` or `transform = ...`) using the methods from `bayestestR`.

DominiqueMakowski:
What's the best approach in your opinion? 

Should we merge this and try to submit it to CRAN first, or bayestestR first and this later?

mattansb:
`bayestestR` should be submitted first - I can disable that example in the vignette so it won't error.

As for testing, I'm not sure how the internals work - is any user specific transformation supported by `modelbased`? Or are the only transformations applied in GLMs?
If the latter, then set up a simple example of binomial / poisson (/both) Bayesian models and their contrasts, ask for a BF... All you really need to test is that it does not fail.

DominiqueMakowski:
I have put some tests here:

https://github.com/easystats/modelbased/blob/87018565631597f730ad08ea33e27313a67b0d27/tests/testthat/test-estimate_contrasts.R#L28-L36

With some specific response "none" and "response"

DominiqueMakowski:
by default, the `transform` arg is set to "response" for estimate_means and for "none" for estimate_contrasts, but that's changeable

mattansb:
I've added specific tests for BFs - *the second test I've added fails*, but as far as I can see this isn't related to code on `bayestestR` end...

DominiqueMakowski:
Thanks, I'll give it a look!

for the record, 99% of the work for `estimate_means()` and contrasts  is done in the [`emmeans_wrapper()`](https://github.com/easystats/modelbased/blob/dev/R/utils_emmeans_wrapper.R) which clumsily attempts to create the right `at` to pass to the `ref_grid` function:

https://github.com/easystats/modelbased/blob/944740e93491aa924fa7841a23f74cf95f393d38/R/utils_emmeans_wrapper.R#L55-L58

PR #97, opened 2021-03-15
edits to vignettes

IndrajeetPatil:
I am wondering if we should be changing the `iris` examples with `palmerpenguins` examples. 

The `penguins` dataset has much more variability and more challenging aspects to it than the `iris` dataset. Agree that we will have to add this package to `Suggests` but I think that shouldn't be a problem for us.

### DIFF
--- a/R/estimate_predicted.R
+++ b/R/estimate_predicted.R
@@ -31,7 +31,7 @@
 #'
 #' # Bayesian models
 #' if (require("rstanarm")) {
-#'   model <- rstanarm::stan_glm(mpg ~ wt, data = mtcars, refresh=0)
+#'   model <- rstanarm::stan_glm(mpg ~ wt, data = mtcars, refresh = 0)
 #'   estimate_response(model)
 #'   estimate_relation(model)
 #' }
@@ -83,12 +83,13 @@ estimate_response <- estimate_prediction
 
   # Get predicted ----------------
   predictions <- insight::get_predicted(model,
-                                        data = data,
-                                        predict = predict,
-                                        ci = ci,
-                                        dispersion_function = "mad",
-                                        interval_function = "hdi",
-                                        ...)
+    data = data,
+    predict = predict,
+    ci = ci,
+    dispersion_function = "mad",
+    interval_function = "hdi",
+    ...
+  )
   out <- as.data.frame(predictions, keep_iterations = keep_iterations)
   out <- cbind(data, out)
 
--- a/inst/WORDLIST
+++ b/inst/WORDLIST
@@ -20,6 +20,7 @@ codecov
 doi
 easystats
 emmeans
+favour
 fdr
 glmmTMB
 hochberg
@@ -29,6 +30,7 @@ joss
 loess
 modelisation
 nd
+polyvalent
 rootSolve
 se
 setosa
--- a/tests/testthat/test-estimate_predicted.R
+++ b/tests/testthat/test-estimate_predicted.R
@@ -8,8 +8,6 @@ osx <- tryCatch({
 })
 
 if (require("testthat") && require("modelbased") && require("gamm4") && require("rstanarm") && require("lme4") && require("glmmTMB") && require("mgcv") && require("MASS") && require("brms") && require("testthat")) {
-
-
   test_that("estimate_link", {
 
     # LMER4
@@ -31,7 +29,7 @@ if (require("testthat") && require("modelbased") && require("gamm4") && require(
 
     # GAMM4
     model <- gamm4::gamm4(Petal.Length ~ Petal.Width + s(Sepal.Length),
-                          random = ~ (1 | Species), data = iris
+      random = ~ (1 | Species), data = iris
     )
     expect_equal(nrow(modelbased::estimate_link(model, length = 3)), 9)
     expect_equal(dim(modelbased::estimate_link(model, include_smooth = FALSE, length = 3)), c(3, 5))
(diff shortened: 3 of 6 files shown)


DominiqueMakowski:
> I am wondering if we should be changing the iris examples with palmerpenguins examples.

I'm aware of all the political controversies and virtue-signalling surrounding the abolition of the iris dataset, and while the idea is in itself good, especially since the penguins dataset offers more options to do statistical demonstrations - which is the main argument for me - <sub>[again, I might be too french and not americanised enough, but it seems to me that revolution is more effective when fought on the streets with guillotines and kings' heads rolling than by changing statistical datasets 😁]</sub> it's super annoying to me to have to install and use an external package (I'm just so used of reprexing `lm(Sepal.Length ~ Species, data=iris)` and old habits die hard 🤷). And I know the iris dataset well, so I'll probably keep using it in the near future out of laziness... So I won't be personally pushing for its replacement, but then if that's something important to you that you want to do then sure :)

IndrajeetPatil:
> I'm aware of all the political controversies and virtue-signalling surrounding the abolition of the iris dataset,

Yeah, me too 😓
  
But, like you clearly state, my concerns are not political/ideological, but rather pedagogical.

I use `iris` dataset in my packages all the time as well and won't be changing those examples. But the difference in `easystats` is that our vignettes are written as beginner-friendly tutorials that try to get the users excited about the package functionality.  But, if I were a beginner, I doubt I'd be much excited by the sepal or petal lengths of a flower, unless I am a botanist 😄 

I will try to play around with the `datasets` package and see if there is another dataset that might be more exciting. This would mean that the users also won't have to install an additional dependency.

mattansb:
What's wrong with mtcars?

IndrajeetPatil:
Too small (only 32 observations)? 

But, I agree, that's definitely a good alternative for now.

DominiqueMakowski:
> What's wrong with mtcars?

and it doesn't have proper factors so it's limited for mixed models etc.

> I will try to play around with the datasets package and see if there is another dataset that might be more exciting.

why not, though I remember looking through a bit, and I never found anything as simple yet flexible as iris (somehow the fact that  some variables are correlated but then that it changes depending on the factor + that there are average differences between the factor is quite nice. The only thing that iris lacks is a binary factor for t-tests and stuff like that). The best dataset would have: One binary outcome 0-1 that one can use to illustrate logistic models as well as a two-levels factor predictors, one factor with 2+ levels to illustrate contrasts, marginal means, random factors, anovas or whatnot, and some numeric variables with interesting relationship. If you find that in `datasets`, you might win the R game 😄 

Though I'm not familiar with it, using penguins could be the best alternative, as long as we explain it to keep it beginner-friendly 🐧

mattansb:
> and it doesn't have proper factors so it's limited for mixed models etc.

``` r
lme4::lmer(mpg ~ factor(cyl) + (1|gear), data = mtcars)
#> boundary (singular) fit: see ?isSingular
#> Linear mixed model fit by REML ['lmerMod']
#> Formula: mpg ~ factor(cyl) + (1 | gear)
#>    Data: mtcars
#> REML criterion at convergence: 157.1612
#> Random effects:
#>  Groups   Name        Std.Dev.
#>  gear     (Intercept) 0.000   
#>  Residual             3.223   
#> Number of obs: 32, groups:  gear, 3
#> Fixed Effects:
#>  (Intercept)  factor(cyl)6  factor(cyl)8  
#>       26.664        -6.921       -11.564  
#> optimizer (nloptwrap) convergence code: 0 (OK) ; 0 optimizer warnings; 1 lme4 warnings
```

It seems silly to load a package for a single dataset.

An alternative is to [simulate some data and add them you our packages](https://easystats.github.io/effectsize/reference/hardlyworking.html) for demo purposes.

DominiqueMakowski:
> lme4::lmer(mpg ~ factor(cyl) + (1|gear), data = mtcars)

okay Mr smartypants

strengejacke:
The penguins dataset is often mentioned due to the fact that you can teach Simpson't paradox - however, this actually can be done with iris as well:

``` r
library(ggplot2)
library(palmerpenguins)
data(iris)
data(penguins)
ggplot(iris, aes(Sepal.Width, Sepal.Length)) +
  stat_smooth(method = "lm", se = FALSE) +
  geom_point(aes(colour = Species)) +
  geom_smooth(aes(colour = Species), method = "lm", se = FALSE)
#> `geom_smooth()` using formula 'y ~ x'
#> `geom_smooth()` using formula 'y ~ x'
```



``` r

ggplot(penguins, aes(bill_length_mm, bill_depth_mm)) +
  stat_smooth(method = "lm", se = FALSE) +
  geom_point(aes(colour = species)) +
  geom_smooth(aes(colour = species), method = "lm", se = FALSE)
#> `geom_smooth()` using formula 'y ~ x'
#> Warning: Removed 2 rows containing non-finite values (stat_smooth).
#> `geom_smooth()` using formula 'y ~ x'
#> Warning: Removed 2 rows containing non-finite values (stat_smooth).
#> Warning: Removed 2 rows containing missing values (geom_point).
```



<sup>Created on 2021-03-17 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

But I see two other pros of penguins:
- unbalanced groups
- missing values

I had often the case that a function works well on the built-in data sets, but failed with "real world examples", where data contains missing values.

I used the penguins dataset in one of workshops when it comes to within/between effects, and to show that mixed models are better than econometric FE models:
[Multilevel Modelling in R - Repeated Measurements.pdf](https://github.com/easystats/modelbased/files/6154452/Multilevel.Modelling.in.R.-.Repeated.Measurements.pdf)

For longitudinal data, we have the "qol_cancer" data in _parameters_: https://easystats.github.io/parameters/articles/demean.html

IndrajeetPatil:
Great points, Daniel!  The unbalanced groups and missing values are indeed good features to have. 

This is why I always test my functions with `iris`, *nice* dataset, and `ggplot2::msleep`, *messy* dataset. Indeed, things that work nicely with `iris` don't always function so well when confronted with messy data.

PR #113, opened 2021-05-31
DESCRIPTION: Remotes: Add missing comma

hsbadr:
This fixes the following error:
```r
Error: Error: Missing commas separating Remotes: 'easystats/bayestestR
    easystats/see'
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -64,7 +64,7 @@ Suggests:
     testthat,
     spelling
 Remotes: 
-    easystats/bayestestR
+    easystats/bayestestR,
     easystats/see
 Encoding: UTF-8
 RoxygenNote: 7.1.1.9001


IndrajeetPatil:
Thanks, but this is already fixed in `master` by now.

DominiqueMakowski:
You're a few seconds too late @hsbadr ^^ but thanks :)

hsbadr:
> You're a few seconds too late @hsbadr ^^ but thanks :)

No worries. I was so lazy to submit a PR for a very minor issue :) Just found your packages and will play with them & see if they support `tidymodels`/`parsnip` models or workflows. They may surely access the fit object, but it'd nice if they do (or can) utilize the preprocessor (e.g., `recipes`) for automatic post-processing (e.g., transform to the original variables  by directly undoing the recipe steps, if required.

DominiqueMakowski:
Tbh we don't offer a super-comprehensive support for tidymodels yet, mostly as we ourselves don't really use that framework. But we are definitely willing to improve on that side, so by all means please don't hesitate to open issues / make PRs here or in the other easystats packages if something is not working as expected!

PR #122, opened 2021-06-14
fix namespace issue

IndrajeetPatil:


R/reexports.R
```diff
@@ -0,0 +1,7 @@
+#' @export
+#' @importFrom effectsize standardize
+effectsize::standardize
```
DominiqueMakowski, line 3: can't we import them directly from where they are?

DominiqueMakowski: Oh you mean this is so that users can use standardize() directly without library(effectsize) is it?

IndrajeetPatil:
You can ignore this. I was just trying to check if this solves the check failures in the master, but it doesn't.

DominiqueMakowski:
but what's going on with this standardzie thing is it because of datawizard?

IndrajeetPatil:
No, nothing to do with `datawizard` since it is currently not being used anywhere in the `easystats` ecosystem.

IndrajeetPatil:
> but what's going on with this standardzie thing

Yeah, that's what I am also trying to figure out.

R/reexports.R
strengejacke: yes, it's a re-export

IndrajeetPatil: Huh, this was the solution but somehow the PR builds still failed. Anyway, doesn’t matter. Seems to be fixed in the master by now.

PR #149, opened 2021-09-24
Add shorter alias for `estimate_*()` functions

bwiernik:
Thinking it might be useful for some folks who complain about too much typing? 

```
export(est_contrasts)
export(est_expect)
export(est_grouplevel)
export(est_link)
export(est_means)
export(est_predict)
export(est_relat)
export(est_response)
export(est_slopes)
```


mattansb:
Yes!!!

bwiernik:
@mattansb dog choices for short names look good?

DominiqueMakowski:
image


I'm not convinced about introducing aliases for something more abstract to "gain" 7 letters ***in the age of tabfill***. Moreover, did you consider all the poor French users for which this looks like `is_response`, `is_slope` ("est" = "is") 😅  

More useful aliases would remove the `estimate_` entirely similarly to how we can drop the `model_` of model_parameters/performance, but here it runs into obivous scoping problems...

@strengejacke watchu think?

DominiqueMakowski:
In general, easystats is fairly consistent in keeping full English names throughout its ecosystem, and I think it has been a really good thing for making the code look transparent to novices/reviewers, so we should try to stick to that

bwiernik:
Yeah, I can see that

strengejacke:
I mean, French folks even say "pétit-deux", thus introducing _longer_ terms than necessary, so it's no surprise that Dom votes against it ;-)

We have not described this in detail in our [code conventions](https://easystats.github.io/easystats/articles/conventions.html), but I think we usually have "full" verb names for functions across easystats (see https://easystats.github.io/easystats/articles/list_of_functions.html).

I personally would follow Dom here and say we shouldn't do .that

DominiqueMakowski:
That said, having shorter aliases could still be considered, at least for some of them?

I'm thinking for instance `modelbased::relation()` would be quite nice, or `modelbased::grouplevel()`. Are there any other that we could easily alias like that?

IndrajeetPatil:
Where do we stand on this PR? There are merge conflicts here that need to be resolved.

bwiernik:
I would say let's just close.

PR #152, opened 2021-10-10
Clarify expectation vs prediction, deprecate estimate_response()

bwiernik:
Following https://github.com/easystats/insight/issues/432, `estimate_response()` was generating a warning, and it isn't clear whether it was an expectation or prediction (IMO). I think deprecating it, making it an alias for `estimate_expectation()` (to match `type="response"` in `predict.glm()`), and pushing users toward expectation/prediction would be good.


strengejacke:
Ok, there seem to be the only these two doc-issue remaining here:

> Warning: prepare_Rd: estimate_response.Rd:65-67: Dropping empty section \details
> checkRd: (5) visualisation_recipe.estimate_predicted.Rd:262-268: Only one \description is allowed

strengejacke:
Now only this one is left:

> Warning:   checkRd: (5) visualisation_recipe.estimate_predicted.Rd:262-268: Only one \description is allowed

DominiqueMakowski:
on it

strengejacke:
> on it

I really don't know which line belongs to which section when the `@` tags are missing ;-) I always explicitly write `@title`, `@description` for that reason...

DominiqueMakowski:
Here the problem was that for some reason it is copying the @description from the original main definition in datawizard 🤔

strengejacke:
Cool, looks ready to merge now!

bwiernik:
Thanks folks!

PR #170, opened 2022-02-16
Patch release 0.7.2, needed for insight update

strengejacke:


strengejacke:
@DominiqueMakowski Once _insight_ is on CRAN, this will break _modelbased_. I have fixed the failing tests, so once you get notified, you can submit _modelbased_ using the version from this PR.

strengejacke:
@DominiqueMakowski I'm not sure if this assignment for `visualization_matrix()` works, you may look at my commits, but I _hope_ that we now assign `visualization_matrix` with `insight::get_datagrid`.

strengejacke:
ok, this doesn't seem to work without a closer look at it, we should take some more time for this...

PR #173, opened 2022-03-05
Correct typo in README.md

moritzkoerber:


IndrajeetPatil:
Thanks, but the `README.md` is actually generated from `README.Rmd`, so you will need to make these corrections to the `.Rmd` file:
https://github.com/easystats/modelbased/blob/master/README.Rmd

moritzkoerber:
> Thanks, but the `README.md` is actually generated from `README.Rmd`, so you will need to make these corrections to the `.Rmd` file: https://github.com/easystats/modelbased/blob/master/README.Rmd

Oops, sorry, should have noticed this directly from the beginning. Fixed!
Thanks for the awesome packages btw!

bwiernik:
Thanks!

Issue #192, opened 2022-07-22
Fix GHA


IndrajeetPatil:
Some tests are failing for only some R version and OS platforms

<img width="841" alt="Screenshot 2022-07-22 at 08 56 50" src="https://user-images.githubusercontent.com/11330453/180381943-ad33bac6-360d-492f-995e-34d21e998636.png">

IndrajeetPatil:
Outdated.

PR #194, opened 2022-07-25
Fix GHA with Ubuntu

etiennebacher:
Related to #192. GSL deps weren't installed for Ubuntu, which created errors later during the package installation.

Possible values for `runner.os` are "Linux", "Windows" or "macOS" (see [here](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context)).


IndrajeetPatil:
Thanks, Etienne!

Issue #163, opened 2021-11-23
adjust argument in ``` estimate_contrasts() ``` is ignored when calculating p-values


docbio78:
It seems there is a bug in estimate_contrasts as the p-values do not change when changing the adjustment method. 
There will be always Tukey-adjusted p-values be reported although the comment in the output says otherwise. 
See for comparison the same output with emmeans. There p-values are different between "tukey" or "fdr".

```
data <- iris
data$Petal.Length_factor <- ifelse(data$Petal.Length < 4.2, "A", "B")
model <- lmer(Sepal.Width ~ Species + (1 | Petal.Length_factor), data = data)

estimate_contrasts(model = model,
+                           lmer.df = "satterthwaite",
+                           adjust = "tukey")
No variable was specified for contrast estimation. Selecting `contrast = "Species"`.
Marginal Contrasts Analysis

Level1     |     Level2 | Difference |        95% CI |   SE |     df |     t |      p
-------------------------------------------------------------------------------------
setosa     | versicolor |       0.87 | [ 0.67, 1.07] | 0.09 | 135.61 | 10.11 | < .001
setosa     |  virginica |       0.80 | [ 0.53, 1.06] | 0.11 | 118.54 |  7.11 | < .001
versicolor |  virginica |      -0.07 | [-0.25, 0.10] | 0.07 | 145.27 | -1.00 | 0.578 

Marginal contrasts estimated at Species
p-value adjustment method: Tukey

> estimate_contrasts(model = model,
+                    lmer.df = "satterthwaite",
+                    adjust = "fdr")
No variable was specified for contrast estimation. Selecting `contrast = "Species"`.
Marginal Contrasts Analysis

Level1     |     Level2 | Difference |        95% CI |   SE |     df |     t |      p
-------------------------------------------------------------------------------------
setosa     | versicolor |       0.87 | [ 0.66, 1.08] | 0.09 | 135.61 | 10.11 | < .001
setosa     |  virginica |       0.80 | [ 0.53, 1.07] | 0.11 | 118.54 |  7.11 | < .001
versicolor |  virginica |      -0.07 | [-0.25, 0.10] | 0.07 | 145.27 | -1.00 | 0.578 

Marginal contrasts estimated at Species
p-value adjustment method: Benjamini & Hochberg (1995)

> # compare to emmeans
> emmeans(model,
+         pairwise ~ Species,
+         lmer.df = "satterthwaite",
+         adjust = "tukey")
$emmeans
[...]

PR #204, opened 2022-08-18
fix: transform and adjust have no effect in estimate_contrast.R

jpasquier:
The `transform` parameter has no effect because it is set to `"none"` in `get_emcontrasts`.

The `adjust` parameter has no effect on the p-values because this parameter is not passed to `get_emcontrasts`.

This PR corrects that.


jpasquier:
solves issue #163

IndrajeetPatil (approved):
Thanks!

the8thday:
although the CRAN version is still not working, bur the giuhub version is OK now, Thanks.

PR #220, opened 2022-12-28
no warn for upcoming insight::get_data refactor

vincentarelbundock:
Minor changes to tests in order to avoid warnings with this upcoming refactor of the `insight::get_data` function:

https://github.com/easystats/insight/pull/691

The issue is that we now recover data from the environment, which is much safer in model with transformations. This poses slight problems with `testthat`, however, since tests are run in weird environments. The tricks are to:

* Avoid using names like `data` and `df` which conflict with lots of internal objects
* Double assign in the data inside test environment: `dat <<- ` Note that this assignment must occur after all transformations to the data.

### DIFF
--- a/tests/testthat/test-estimate_contrasts.R
+++ b/tests/testthat/test-estimate_contrasts.R
@@ -1,7 +1,8 @@
 if (require("logspline") && require("rstanarm") && require("lme4") && require("emmeans")) {
   test_that("estimate_contrasts - Frequentist", {
     # One factor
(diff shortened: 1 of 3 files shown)


vincentarelbundock:
I think this should be merged because the failures are already present in `main` and are unrelated to this PR.

I don't have merge right on this repo, so I can't do this myself.

IndrajeetPatil (approved):

IndrajeetPatil:
Thanks!

PR #227, opened 2023-05-08
Add effect size (Cohen's d) to `estimate_contrasts`

rempsyc:
As discussed in easystats/report#372

``` r
library(modelbased)

model <- lm(Sepal.Width ~ Species, data = iris)
estimate_contrasts(model)
#> No variable was specified for contrast estimation. Selecting `contrast = "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |         95% CI |   SE | t(147) |      p | Cohen's d | Cohens_d 95% CI
#> ------------------------------------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.49,  0.82] | 0.07 |   9.69 | < .001 |      1.89 |  [ 1.41,  2.36]
#> setosa     |  virginica |       0.45 | [ 0.29,  0.62] | 0.07 |   6.68 | < .001 |      1.29 |  [ 0.86,  1.72]
#> versicolor |  virginica |      -0.20 | [-0.37, -0.04] | 0.07 |  -3.00 | 0.003  |     -0.64 |  [-1.04, -0.24]
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)

model <- lm(Sepal.Width ~ Species * Petal.Width, data = iris)
estimate_contrasts(model)
#> No variable was specified for contrast estimation. Selecting `contrast = "Species"`.
#> NOTE: Results may be misleading due to involvement in interactions
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |        95% CI |   SE | t(144) |      p | Cohen's d | Cohens_d 95% CI
#> -----------------------------------------------------------------------------------------------------------
#> setosa     | versicolor |       1.59 | [ 0.64, 2.54] | 0.39 |   4.04 | < .001 |      1.89 |  [ 1.41,  2.36]
#> setosa     |  virginica |       1.77 | [ 0.77, 2.78] | 0.41 |   4.29 | < .001 |      1.29 |  [ 0.86,  1.72]
#> versicolor |  virginica |       0.18 | [-0.17, 0.54] | 0.15 |   1.27 | 0.205  |     -0.64 |  [-1.04, -0.24]
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)

data <- mtcars
data$cyl <- as.factor(data$cyl)
data$am <- as.factor(data$am)
model <- rstanarm::stan_glm(mpg ~ cyl * am, data = data, refresh = 0)
estimate_contrasts(model)
#> No variable was specified for contrast estimation. Selecting `contrast = "cyl"`.
#> NOTE: Results may be misleading due to involvement in interactions
#> Marginal Contrasts Analysis
#> 
#> Level1 | Level2 | Difference |        95% CI |     pd | % in ROPE | Cohen's d | Cohens_d 95% CI
#> -----------------------------------------------------------------------------------------------
#> cyl4   |   cyl6 |       5.62 | [2.31,  8.85] | 99.95% |        0% |      1.88 |    [0.72, 3.01]
#> cyl4   |   cyl8 |      10.23 | [7.06, 13.48] |   100% |        0% |      3.26 |    [2.02, 4.48]
#> cyl6   |   cyl8 |       4.67 | [1.34,  8.00] | 99.62% |        0% |      2.05 |    [0.91, 3.14]
#> 
#> Marginal contrasts estimated at cyl
```

<sup>Created on 2023-05-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Though it will skip it edge cases where the following arguments are not `NULL`: `contrast`, `fixed`, and `at`, because those made it hard to compute Cohen's _d_, which right now fetches the original data to compute it. I don't know if the contrasts modified data is available anywhere so we could calculate proper effect sizes for those cases?

For now though I guess it is still a minor improvement to no effect sizes at all.

---

- [x] Use `emmeans::eff_size()`
- [x] Use `bootES::bootES()`
- [x] ~~Derive ES from the marginal statistics~~ (not sure what Dom meant by this, maybe this method is addressed via the other methods?)
- [ ] Refit the model with just the variables used in the contrasts and get sigma from that (easystats/report#372)
- [x] Use ES from easystats/effectsize#351
- [x] ~~Use ES from easystats/effectsize#372~~ (looked into 372 but cannot figure it out...)


DominiqueMakowski:
thanks! and do add yourself in the authors desc

R/estimate_contrasts.R
```diff
@@ -135,6 +135,31 @@ estimate_contrasts <- function(model,
...
+      dat <- datawizard::data_select(dat, c(resp, info$contrast))
+
+      list.dat <- lapply(seq_len(nrow(contrasts)), function(i) {
+        log.vec <- which(dat[[info$contrast]] == unlist(info$misc$orig.grid)[[i]])
+        dat.temp <- datawizard::data_filter(dat, log.vec)
+      })
+      list.dat <- stats::setNames(list.dat, unique(unlist(level_cols)))
+
+      eff <- lapply(seq_len(nrow(contrasts)), function(i) {
+        effectsize::cohens_d(x = list.dat[[contrasts$Level1[i]]][[resp]],
```
DominiqueMakowski, line 153: I wonder if getting raw Cohen's *d* like that from the raw data instead of relying on the marginal statistics isn't misleading though 🤔 @mattansb @bwiernik what do you think?

mattansb: Neither option is ideal (see discussion here: https://github.com/easystats/effectsize/issues/212, and note that `cohens_d()` has a `paired = ` argument).

I think you're looking for this https://github.com/easystats/effectsize/issues/351?

rempsyc: My approach to calculating effect sizes in [`rempsyc::nice_contrasts`](https://rempsyc.remi-theriault.com/reference/nice_contrasts.html) is to use [`bootES::bootES`](https://github.com/dgerlanc/bootES/blob/main/R/bootES.R), which allows specifying contrasts (and a robust version of Cohen's _d_, and other cool things).

For `modelbased`, I wanted to stay within the easystats ecosystem to avoid adding another dependency, so I went for the regular Cohen's _d_ for the short-term, forgetting about easystats/effectsize#351. But if I understand correctly, #351 is not ready.

So that leaves us four choices for the _short-term_:

1. Stick with Cohen's _d_, even if it is not ideal
2. Derive it from the marginal statistics, like Dom suggested
3. Use `bootES::bootES` like in `rempsyc`
4. Merge #351 so I can use that.

Obviously no 4 is best, but I don't know what is missing so we can merge it? If we have to wait for ever for it to be fixed (it's been 2 years already), I'd rather have a short-term solution. So what do you want me to do short-term @mattansb @DominiqueMakowski ?

bwiernik: We can compute legit cohens d from the coefficients/marginal contrasts by adjusting the sigma to add back any other covariate variance that has been pulled out

mattansb: For anything other than a linear regression with a single factor predictor, computing a standardized difference requires decisions (do we adjust for the covariances or not? Do we standardize by the residual sd or do we also account for level-2 variance(s)?).
There are not trivial decisions, so I suggest *not* including these by default (or at all).

Maybe let's be part of the movement that encourages people to think about their X/Y scales? The movement to mean people off of standardized effect sizes?

mattansb: (saw this, and will reply once the situation here gets back to normal ✌️)

rempsyc: > Maybe let's be part of the movement that encourages people to think about their X/Y scales? The movement to mean people off of standardized effect sizes?

I’m all for using original scale units when those are meaningful. But in my department for example, almost everyone relies exclusively on Likert scales, many questionnaire also use a different number of response options. In this context, do you still think it makes more sense to use non-standardized scores?

> (do we adjust for the covariances or not? Do we standardize by the residual sd or do we also account for level-2 variance(s)?) There are not trivial decisions, so I suggest not including these by default (or at all).

I’d have to agree with Dom here. A lot of people argue that contrast analysis is superior to plain _t_ tests. But with _t_ tests at least you can get Cohen’s _d_. So it seems like there is no point to do contrasts analysis if there is not yet a satisfying solution to get the effect size.

So what will people do? Take the next best option. My colleagues are likely to simply attempt to compute Cohen’s _d_ on the groups directly, even if this is mistaken with contrasts. Slightly more savvy people like me might think of using `bootES::bootES()` or `emmeans::eff_size()`. I guess that’s the best currently implemented options.

So I think providing a reasonable default with some decisions taken for the users, yes, it will actually prevent greater harm (people doing Cohen's _d_ by hand because no effect size is provided).

I think it's not always reasonable to expect most people to understand stats at such a """deep""" level e.g., for that kind of decision-making. These are not trivial decisions, granted. But I'm pretty sure nobody at my department would even know what decision to take there. They wouldn't even know what it means or the implications. I mean I can't blame them. They have to become experts as therapists and clinicians, at the theories of psychology, at methodology, at statistical analysis, at writing, at public speaking, at managing others, at marketing themselves, etc., AND they might have to work outside school, or they might be volunteering in student clubs or sports or whatnot. That's a lot to ask :)

So if we can't even agree on the effect size here, imagine those regular non-savvy students! Other fields frequently have a dedicated statistician to their research team. But (sadly?), in psychology, this is not the norm. So I think that's why people like `easystats` so much. It makes it easy to get "good enough" outputs by default. So as long as our decision-making is documented and can be overwritten, I think we're OK.

> Is whatever bootES() provides legit?

Well, after double-checking, even though it supports contrasts, it works with the original data frame and the contrasts provided, not with the `lm` model, so I think it would not take into account all the other stuff specified in the model like covariates, etc.

> We can compute legit cohens d from the coefficients/marginal contrasts by adjusting the sigma to add back any other covariate variance that has been pulled out

Ultimately, I think Brenton’s solution would be the most realistic: best compromise of effort and output acceptability. I just don’t understand what it means and how I could implement it myself sadly 😬

mattansb: > I hear that, but I still think adding an option to get reasonable effect size (by default) would be good.

If you review the lit, I think you will find that people almost never report Cohen's d for contrasts, and when they do it is not clear what standardization method they are using, so there is no "reasonable" option that is required or available. If ***we*** decide to give some measure of effect size, we will be setting a *new* default.

> Now as to what a "reasonable" ES is in that case, well I would naively say "whatever is the most natural extension of a traditional Cohen's d".

Here, then, I think I disagree with @bwiernik - the natural extension would use the residual sigma for the standardization, giving a standardized difference between conditional population means.

> I don't know what the answer there, but returning to the initial purpose of effectsize it was mostly to have something to give an intuitive "feel" of effect magnitude. [...]

I completely disagree with this whole paragraph 😅. If you want to give people an intuitive feel for the size of their contrasts, it is SUPER IMPORTANT that you compute a proper effect size! If we want to give people what they *want* from contrasts, I'd argue they only want the sign (positive of negative difference) and a p-value.... If ***we*** decide to give some measure of effect size, we are responsible for giving *a good one*.

---

> I’m all for using original scale units when those are meaningful. But in my department for example, almost everyone relies exclusively on Likert scales, many questionnaire also use a different number of response options. In this context, do you still think it makes more sense to use non-standardized scores?

Absolutely! Within the context of a 1-7 scale, a change of 1.2 units is much more understandable to me than "a change of 0.4 pooled SDs".

> I’d have to agree with Dom here. A lot of people argue that contrast analysis is superior to plain t tests. But with t tests at least you can get Cohen’s d. So it seems like there is no point to do contrasts analysis if there is not yet a satisfying solution to get the effect size.

You can say the same thing about regression - regression is superior to a correlation, yet the standardized coefficients are harder to interpret (can be larger than 1, for example), so that makes the method less useful? No, it just means that simple problems have simple solutions.

But here we can agree - if you are doing a one-way (between subjects) ANOVA instead of 16 t-tests, it is trivial to compute Cohen's d for the contrasts - the estimated difference divided by the pooled sd.

... I personally prefer people do Cohen's d by hand. I see no reason not to do that. If your design is simple enough, the only difference between a t-test and contrasts is the SE, the raw estimate will be the same.

---

The ideal solution to me would be:
1. No standardized effect sizes by default.
2. Have **several** methods for standardized effect sizes (essentially wrappers around something like `emmeans::eff_size()` to make use of it easy) for how one can standardize their estimates + a vignette explaining the differences.

This can cover a range of methods, such as https://github.com/easystats/effectsize/issues/351 or even  https://github.com/easystats/effectsize/issues/372 for rm data / mixed models.

This would be an effort I can get behind.

rempsyc: I think that's reasonable and I can get behind your suggestion too. So then let's not rush things just to have something short-term (so I'll close this PR since that was the original objective). We can probably start working on implementing these solutions bit by bit.

> No standardized effect sizes by default.

Would that mean _no_ effect sizes by default or _non-standardized_ effect sizes by default? And what does this look like? I suppose it's not just the mean difference since like you were saying we may need to adjust for other stuff in the model like covariates and other predictors? So just adjusted mean difference?

DominiqueMakowski: > - No standardized effect sizes by default.
> Have several methods for standardized effect sizes (essentially wrappers around something like emmeans::eff_size() to make use of it easy) for how one can standardize their estimates + a vignette explaining the differences.

That sounds fair

mattansb: > I think this is just the Difference column in the current output haha ("marginal means"), that's the (non-standardized) effect size.

Yup.

YAY!

rempsyc:
Current reprexes:

``` r
library(modelbased)

model <- lm(Sepal.Width ~ Species, data = iris)

estimate_contrasts(model)
#> No variable was specified for contrast estimation. Selecting `contrast = "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |         95% CI |   SE | t(147) |      p
#> ------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.49,  0.82] | 0.07 |   9.69 | < .001
#> setosa     |  virginica |       0.45 | [ 0.29,  0.62] | 0.07 |   6.68 | < .001
#> versicolor |  virginica |      -0.20 | [-0.37, -0.04] | 0.07 |  -3.00 | 0.003 
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)

estimate_contrasts(model, effectsize = "cohens_d")
#> No variable was specified for contrast estimation. Selecting `contrast = "Species"`.
#> Unsupported effect size 'cohens_d', returning none.
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |         95% CI |   SE | t(147) |      p
#> ------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.49,  0.82] | 0.07 |   9.69 | < .001
#> setosa     |  virginica |       0.45 | [ 0.29,  0.62] | 0.07 |   6.68 | < .001
#> versicolor |  virginica |      -0.20 | [-0.37, -0.04] | 0.07 |  -3.00 | 0.003 
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)

estimate_contrasts(model, effectsize = "emmeans")
#> No variable was specified for contrast estimation. Selecting `contrast = "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |         95% CI |   SE | t(147) |      p | effect_size |      es 95% CI
#> -------------------------------------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.49,  0.82] | 0.07 |   9.69 | < .001 |        1.94 | [ 1.48,  2.39]
#> setosa     |  virginica |       0.45 | [ 0.29,  0.62] | 0.07 |   6.68 | < .001 |        1.34 | [ 0.91,  1.76]
#> versicolor |  virginica |      -0.20 | [-0.37, -0.04] | 0.07 |  -3.00 | 0.003  |       -0.60 | [-1.00, -0.20]
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)

set.seed(100)
estimate_contrasts(model, effectsize = "bootES")
#> No variable was specified for contrast estimation. Selecting `contrast = "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |         95% CI |   SE | t(147) |      p | cohens.d | cohens.d 95% CI
#> -----------------------------------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.49,  0.82] | 0.07 |   9.69 | < .001 |     1.94 |  [ 1.52,  2.39]
#> setosa     |  virginica |       0.45 | [ 0.29,  0.62] | 0.07 |   6.68 | < .001 |     1.34 |  [ 0.82,  1.72]
#> versicolor |  virginica |      -0.20 | [-0.37, -0.04] | 0.07 |  -3.00 | 0.003  |    -0.60 |  [-0.92, -0.22]
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)

set.seed(100)
estimate_contrasts(model, effectsize = "bootES", bootES_type = "akp.robust.d")
#> No variable was specified for contrast estimation. Selecting `contrast = "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |         95% CI |   SE | t(147) |      p | akp.robust.d | akp.robust.d 95% CI
#> -------------------------------------------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.49,  0.82] | 0.07 |   9.69 | < .001 |         1.88 |      [ 1.42,  2.49]
#> setosa     |  virginica |       0.45 | [ 0.29,  0.62] | 0.07 |   6.68 | < .001 |         1.37 |      [ 0.92,  1.78]
#> versicolor |  virginica |      -0.20 | [-0.37, -0.04] | 0.07 |  -3.00 | 0.003  |        -0.51 |      [-0.90, -0.17]
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)
```

<sup>Created on 2023-07-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

For the `emmeans` effect size, I just called it `effect_size`, because I don't know if we can call it Cohen's _d_ (or something else).

I updated the to-do list at the top post with the missing effect sizes

R/estimate_contrasts.R
```diff
@@ -11,9 +11,32 @@
...
+#' @section Effect Size: By default, `estimate_contrasts` reports no standardized effect size
+#' on purpose. Should one request one, some things are to keep in mind. As the
+#' authors of `emmeans` write, "There is substantial disagreement among
+#' practitioners on what is the appropriate sigma to use in computing effect
+#' sizes; or, indeed, whether any effect-size measure is appropriate for some
+#' situations. The user is completely responsible for specifying appropriate
+#' parameters (or for failing to do so)."
+#'
+#' In particular, effect size methods `"emmeans"` and `"bootES"` do not correct
+#' for covariates in the model, so should probably only be used when there is
```
mattansb, line 33: I don't know about bootES, but emmeans does return adjusted (corrected?) effect sizes.

R/estimate_contrasts.R
```diff
@@ -11,9 +11,32 @@
...
+#' just one categorical predictor (with however many levels). If there are
+#' multiple predictors or any covariates, it is important to re-compute sigma
+#' adding back in the response variance associated with the variables that
```
mattansb, line 36: I know @bwiernik says this, but I am not convinced this -- as a blanket statement -- is true.
Adjusting predictions in multiple regression is something we do all the time - why is it valid for the means but not for the variance?

rempsyc:
Alright, I added a new method, `"marginal"`, based on Brenton's formula. That said,

``` r
library(modelbased)

mtcars2 <- mtcars
mtcars2$cyl <- as.factor(mtcars2$cyl)

model <- lm(mpg ~ cyl + wt * hp, mtcars2)
```

*d* values from method “marginal”,

``` r
estimate_contrasts(model, effectsize = "marginal")
#> No variable was specified for contrast estimation. Selecting `contrast = "cyl"`.
#> Marginal Contrasts Analysis
#> 
#> Level1 | Level2 | Difference |        95% CI |   SE | t(26) |      p | marginal_d
#> ---------------------------------------------------------------------------------
#> cyl4   |   cyl6 |       1.26 | [-2.55, 5.07] | 1.49 |  0.85 | > .999 |       0.19
#> cyl4   |   cyl8 |       1.45 | [-3.83, 6.74] | 2.06 |  0.70 | > .999 |       0.22
#> cyl6   |   cyl8 |       0.20 | [-3.64, 4.03] | 1.50 |  0.13 | > .999 |       0.03
#> 
#> Marginal contrasts estimated at cyl
#> p-value adjustment method: Holm (1979)
```

differs a lot from emmeans,

``` r
estimate_contrasts(model, effectsize = "emmeans")
#> No variable was specified for contrast estimation. Selecting `contrast = "cyl"`.
#> Marginal Contrasts Analysis
#> 
#> Level1 | Level2 | Difference |        95% CI |   SE | t(26) |      p | effect_size |     es 95% CI
#> --------------------------------------------------------------------------------------------------
#> cyl4   |   cyl6 |       1.26 | [-2.55, 5.07] | 1.49 |  0.85 | > .999 |        0.57 | [-0.83, 1.97]
#> cyl4   |   cyl8 |       1.45 | [-3.83, 6.74] | 2.06 |  0.70 | > .999 |        0.66 | [-1.27, 2.60]
#> cyl6   |   cyl8 |       0.20 | [-3.64, 4.03] | 1.50 |  0.13 | > .999 |        0.09 | [-1.31, 1.49]
#> 
#> Marginal contrasts estimated at cyl
#> p-value adjustment method: Holm (1979)
```

Although they seem consistent. Did I get the formula right? I used

``` r
R2_cov <- summary(model)$r.squared
d_adj <- contrasts$t * contrasts$SE / sigma(model) * sqrt(1 - R2_cov)
```

<sup>Created on 2023-07-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

rempsyc:
Method `marginal` (should we use a different name, like `marginal_d` or `conditional_d`?) for now also does not have confidence intervals. I know that easystats/effectsize#351 has them in the large function, but I wasn't sure if I should implement them here in `modelbased` or wait for it to be formally implemented in `effectsize` (since the issue belongs to that repo). At least for now we have an easy way to get the marginal/conditional d-value itself.

R/estimate_contrasts.R
rempsyc: Thanks, I removed `emmeans` from that paragraph and only left `bootES`.

R/estimate_contrasts.R
rempsyc: Ok, I added "Some believe that", any suggestion as to how else I should change that paragraph? I suppose the intention is to try to warn users about the different risks involved...

``` r
#' In particular, effect size method `"bootES"` does not correct
#' for covariates in the model, so should probably only be used when there is
#' just one categorical predictor (with however many levels). Some believe that
#' if there are multiple predictors or any covariates, it is important to
#' re-compute sigma adding back in the response variance associated with the
#' variables that aren't part of the contrast.
```

R/estimate_contrasts.R
rempsyc: Actually the reason why I included `emmeans` in this paragraph is that in #easystats/effectsize351, @bwiernik writes:

> Method 2: Standardize using the MSE (sigma). This is what `emmeans::eff_size()` does. This works when the contrasts are the only predictors in the model, but not when there are covariates. The response variance accounted for by the covariates should not be removed from the SD used to standardize. Otherwise, d will be overestimated as shown above. I can pull some references from Larry Hedges about this.

So I'm not sure I understand correctly whether `emmeans` correct (correctly) for covariates or not?

R/estimate_contrasts.R
mattansb: Thanks (:
It is perhaps more efficient to replace this whole section with "DON'T" 😜

R/estimate_contrasts.R
mattansb: Well that depends on what you think the correct correction is 😅
emmeans uses the conditional SD, whereas Brenton is in the opinion that the marginal SD should be used (:

rempsyc:
@bwiernik if you think this looks good after reading my previous three comments above, then I think we'll be ready to think about merging this. Anybody else you would like me to implement (e.g., see from list from original post)?

R/estimate_contrasts.R
```diff
@@ -11,9 +11,49 @@
...
+#' `effectsize = "emmeans"` uses [emmeans::eff_size] with
+#' `sigma = stats::sigma(model)`, `edf = stats::df.residual(model)` and
+#' `method = "identity")`. This standardizes using the MSE (sigma). Some believe
+#' this works when the contrasts are the only predictors in the model, but not
+#' when there are covariates. The response variance accounted for by the
+#' covariates should not be removed from the SD used to standardize. Otherwise,
+#' _d_ will be overestimated.
+#'
+#' `effectsize = "marginal"` uses the following formula to compute effect
+#' size: `d_adj <- t * se_b / sigma * sqrt(1 - R2_cov)`. This standardized
```
mattansb, line 49: Note that `t * se_b` is equal to the raw difference - so this can be rewritten as `difference * (1- R2)/ sigma`, which I think is more clear.

R/estimate_contrasts.R
```diff
@@ -11,9 +11,49 @@
...
+#' so)."
+#'
+#' In particular, effect size method `"bootES"` does not correct
+#' for covariates in the model, so should probably only be used when there is
+#' just one categorical predictor (with however many levels). Some believe that
+#' if there are multiple predictors or any covariates, it is important to
+#' re-compute sigma adding back in the response variance associated with the
+#' variables that aren't part of the contrast.
+#'
+#' `effectsize = "emmeans"` uses [emmeans::eff_size] with
```
mattansb, line 40: I'm not sure about this name. `emmeans`'s `eff_size()` can be used with many different sigmas (see next comment)... How about "partial" or something like that?

R/estimate_contrasts.R
mattansb: This also means that `emmeans::eff_size()` can be used here with `sigma = stats::sigma(model) / (1 - R2_cov)` (and some reasonable `edf`)

R/estimate_contrasts.R
rempsyc: Like name it `partial_effect_size`?

R/estimate_contrasts.R
rempsyc: Ok, one thing to note, is that previously I hardcoded sigma as as `sigma = stats::sigma(model)`. Would it be preferable to add a new argument `sigma` with that as the default, but then people can specify a different one when using the `emmeans` method? And Should this argument also apply to Brenton's `marginal` method? (again, instead of hard coding it as `sigma(model)`)

rempsyc: Similarly, in the `emmeans` method, `edf` is currently hardcoded as `edf = stats::df.residual(model)`, and method  as `method = "identity"`. But I can add arguments for those as well (I'm always unsure whether the ellipsis three dots would work when they can be used in several functions with similar argument names, it would probably create conflict for `method` for example).

mattansb: Well, in that case you're just making a wrapper around `eff_size`, yeah?

R/estimate_contrasts.R
mattansb: `"partial d"`, maybe?

R/estimate_contrasts.R
rempsyc: Yes. But I don't know what we should do (wrapper around `eff_size` with some defaults, or fixed values). Do you have a preference?

mattansb: nothing 😜🤷‍♂️

DominiqueMakowski:
@rempsyc it would be potentially nice to have something like this in for 1.0, what were the main blockers here?

rempsyc:
I think I just needed approval to merge, but now it's been so long I have to merge the conflicts first

DominiqueMakowski:
> I think I just needed approval to merge

rempsyc:
> @DominiqueMakowski if you think this looks good after reading my previous three comments above, then I think we'll be ready to think about merging this. Anything else you would like me to implement (e.g., see from list from original post)?

image

R/estimate_contrasts.R
```diff
@@ -10,9 +10,49 @@
 #' comparisons. Can be one of `"holm"` (default), `"tukey"`, `"hochberg"`,
 #' `"hommel"`, `"bonferroni"`, `"BH"`, `"BY"`, `"fdr"` or `"none"`. See the
 #' p-value adjustment section in the `emmeans::test` documentation.
+#' @param effectsize Desired measure of standardized effect size, one of "none"
+#' (default), "emmeans", "marginal", or "bootES".
+#' @param bootES_type Specifies the type of effect-size measure to
```
DominiqueMakowski, line 15: I would not document these in the main interface (to not alourdir down the list of arguments) and allow their change via Kwargs, and mention them within the details section

rempsyc: do you want this merged before submitting the paper?

strengejacke:
@mattansb Don't we have corresponding functions in the effectsize-package? What about supporting _marginaleffects_, is this possible? That's our new default backend...

mattansb:
There isn't any function for this in `effectsize` - for anything by the simplest model (a one-way anova: linear, no repeated measures, no covariates), there are so many choices to be made and no good defaults (or at the very least, no well established norms) - what variance to use for standardize, account for covariates, etc... I mean, just look at [the mess going on for a simple paired t-test...](https://easystats.github.io/effectsize/reference/repeated_measures_d.html).

I ***strongly*** suggest only supporting these for models _without_ random effects.

Also these are only relevant for families with a variance parameter - so essentially only for gaussian models.

strengejacke:
So we end up just using `effectsize::cohens_d()` for a limited range of models?

mattansb:
No, I think it makes sense for the feature @rempsyc build to be here - it gives standardized differences that are based on a model's parameters - or _modelbased_, if you will (:

`effectsize::cohens_d()` gives many options, but only for a two sample and one sample case (converting a paired-sample to a one-sample case of differences).

R/estimate_contrasts.R
DominiqueMakowski: if possible why not

R/estimate_contrasts_effecsize.R
```diff
@@ -0,0 +1,61 @@
...
+    # d_adj <- contrasts$t * contrasts$SE / sigma(model) * sqrt(1 - R2)
+    # New: d_adj <- difference * (1- R2)/ sigma
+    R2 <- summary(model)$r.squared
+    d_adj <- contrasts$Difference * (1 - R2) / sigma(model)
+    contrasts <- cbind(contrasts, marginal_d = d_adj)
+
+  } else if (effectsize == "bootES") {
+    if (bootstraps < 500) {
+      insight::format_alert("Number of bootstraps probably too low. Consider increasing it.")
+    }
```
DominiqueMakowski, line 28:
```suggestion
```

DominiqueMakowski: Let's assume the user has good reason to explicitly set what stuff

R/estimate_contrasts_effecsize.R
```diff
@@ -0,0 +1,61 @@
...
+        group.col = group,
+        contrast = x,
+        effect.type = bootES_type
+      )
+      y <- as.data.frame(summary(y))
+    })
+
+    eff <- do.call(rbind, es.lists)
+    eff <- eff[1:3]
+    names(eff) <- c(bootES_type, paste0(bootES_type, "_CI_low"), paste0(bootES_type, "es_CI_high"))
```
DominiqueMakowski, line 56:
```suggestion
    names(eff) <- c(bootES_type, paste0(bootES_type, "_CI_low"), paste0(bootES_type, "_CI_high"))
```

DominiqueMakowski (changes requested):

rempsyc:
> I strongly suggest only supporting these for models without random effects.

So here's what's left to do:

- [ ] check whether the model includes random effects (with `insight`?)
- [ ] if yes, output without effect size with a warning that this is not supported
- [ ] right now `.estimate_contrasts_effecsize` isn't actually used (so no effect size is used), so integrate back

rempsyc:
How can I access the `marginaleffects` contrast grid as in `emmeans`?

```r
estimated@misc$con.coef
                       setosa versicolor virginica
setosa - versicolor         1         -1         0
setosa - virginica          1          0        -1
versicolor - virginica      0          1        -1
```

If we want to make the `bootES` effectsize method compatitble with `marginaleffects`.

rempsyc:
Current reprex:

``` r
library(modelbased)

model <- lm(Sepal.Width ~ Species, data = iris)

estimate_contrasts(model, effectsize = "emmeans", backend = "emmeans")
#> No variable was specified for contrast estimation. Selecting `contrast =
#>   "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     | Level2     | Difference |         95% CI |   SE | t(147) |      p
#> ------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.52,  0.79] | 0.07 |   9.69 | < .001
#> setosa     | virginica  |       0.45 | [ 0.32,  0.59] | 0.07 |   6.68 | < .001
#> versicolor | virginica  |      -0.20 | [-0.34, -0.07] | 0.07 |  -3.00 |  0.003
#> 
#> Level1     | partial_d |      es 95% CI
#> ---------------------------------------
#> setosa     |      1.94 | [ 1.48,  2.39]
#> setosa     |      1.34 | [ 0.91,  1.76]
#> versicolor |     -0.60 | [-1.00, -0.20]
#> 
#> Variable predicted: Sepal.Width
#> Predictors contrasted: Species
#> p-values are uncorrected.

estimate_contrasts(model, effectsize = "marginal", backend = "emmeans")
#> No variable was specified for contrast estimation. Selecting `contrast =
#>   "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     | Level2     | Difference |         95% CI |   SE | t(147) |      p | marginal_d
#> -------------------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.52,  0.79] | 0.07 |   9.69 | < .001 |       1.16
#> setosa     | virginica  |       0.45 | [ 0.32,  0.59] | 0.07 |   6.68 | < .001 |       0.80
#> versicolor | virginica  |      -0.20 | [-0.34, -0.07] | 0.07 |  -3.00 |  0.003 |      -0.36
#> 
#> Variable predicted: Sepal.Width
#> Predictors contrasted: Species
#> p-values are uncorrected.

estimate_contrasts(model, effectsize = "bootES", backend = "emmeans")
#> No variable was specified for contrast estimation. Selecting `contrast =
#>   "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     | Level2     | Difference |         95% CI |   SE | t(147) |      p
#> ------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.52,  0.79] | 0.07 |   9.69 | < .001
#> setosa     | virginica  |       0.45 | [ 0.32,  0.59] | 0.07 |   6.68 | < .001
#> versicolor | virginica  |      -0.20 | [-0.34, -0.07] | 0.07 |  -3.00 |  0.003
#> 
#> Level1     | cohens.d | cohens.d 95% CI
#> ---------------------------------------
#> setosa     |     1.94 |  [ 1.53,  2.43]
#> setosa     |     1.34 |  [ 0.91,  1.76]
#> versicolor |    -0.60 |  [-1.04, -0.30]
#> 
#> Variable predicted: Sepal.Width
#> Predictors contrasted: Species
#> p-values are uncorrected.

estimate_contrasts(model,
                   effectsize = "bootES",
                   bootES_type = "akp.robust.d",
                   backend = "emmeans"
)
#> No variable was specified for contrast estimation. Selecting `contrast =
#>   "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     | Level2     | Difference |         95% CI |   SE | t(147) |      p
#> ------------------------------------------------------------------------------
#> setosa     | versicolor |       0.66 | [ 0.52,  0.79] | 0.07 |   9.69 | < .001
#> setosa     | virginica  |       0.45 | [ 0.32,  0.59] | 0.07 |   6.68 | < .001
#> versicolor | virginica  |      -0.20 | [-0.34, -0.07] | 0.07 |  -3.00 |  0.003
#> 
#> Level1     | akp.robust.d | akp.robust.d 95% CI
#> -----------------------------------------------
#> setosa     |         1.88 |      [ 1.41,  2.53]
#> setosa     |         1.37 |      [ 0.90,  1.81]
#> versicolor |        -0.51 |      [-0.94, -0.15]
#> 
#> Variable predicted: Sepal.Width
#> Predictors contrasted: Species
#> p-values are uncorrected.

estimate_contrasts(model, effectsize = "marginal", backend = "marginaleffects")
#> We selected `contrast=c("Species")`.
#> Marginal Contrasts Analysis
#> 
#> Level1     | Level2     | Difference |   SE |         95% CI | t(147) |      p | marginal_d
#> -------------------------------------------------------------------------------------------
#> versicolor | setosa     |      -0.66 | 0.07 | [-0.79, -0.52] |  -9.69 | < .001 |      -1.16
#> virginica  | setosa     |      -0.45 | 0.07 | [-0.59, -0.32] |  -6.68 | < .001 |      -0.80
#> virginica  | versicolor |       0.20 | 0.07 | [ 0.07,  0.34] |   3.00 |  0.003 |       0.36
#> 
#> Variable predicted: Sepal.Width
#> Predictors contrasted: Species
#> p-values are uncorrected.

estimate_contrasts(model, effectsize = "emmeans", backend = "marginaleffects")
#> We selected `contrast=c("Species")`.
#> Error: `effectsize = emmeans` only possible with `backend = emmeans`

estimate_contrasts(model, effectsize = "bootES", backend = "marginaleffects")
#> We selected `contrast=c("Species")`.
#> Error in estimated@misc: no applicable method for `@` applied to an object of class "marginaleffects_contrasts"

sleepstudy <- lme4::sleepstudy
sleepstudy$Days_factor <- cut(sleepstudy$Days, breaks = 3, labels = c("Low", "Medium", "High"))
model_random_effects <- lme4::lmer(Reaction ~ Days_factor + (1 | Subject), data = sleepstudy)

estimate_contrasts(model_random_effects, effectsize = "marginal")
#> We selected `contrast=c("Days_factor")`.
#> Error: We strongly recommend not using the `effectsize` argument with models
#>   containing random effects.
```

<sup>Created on 2025-02-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

rempsyc:
Now I really have to focus on my job application due next week which I haven't started yet 😬 but I trust you guys if you want to merge and move forward

mattansb:
@rempsyc you should also check that the model is Gaussian.

rempsyc:
image

💀

rempsyc:
``` r
library(insight)
#> Warning: package 'insight' was built under R version 4.4.2

model <- lm(Sepal.Width ~ Species, data = iris)

is_gaussian(model)
#> Error in is_gaussian(model): could not find function "is_gaussian"
# 🥲

get_family(model)$family == "gaussian"
#> [1] TRUE

# or
model_info(model)$family == "gaussian"
#> [1] TRUE
```

<sup>Created on 2025-03-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

rempsyc:
> @rempsyc you should also check that the model is Gaussian.

Fixed!

``` r
library(modelbased)

iris$IsSetosa <- as.integer(iris$Species == "setosa")

model <- glm(IsSetosa ~ Species + Sepal.Length, family = binomial, data = iris)

insight::get_family(model)$family == "Gaussian"
#> [1] FALSE

estimate_contrasts(model, effectsize = "emmeans", backend = "emmeans")
#> No variable was specified for contrast estimation. Selecting `contrast =
#>   "Species"`.
#> Error: We strongly recommend not using the `effectsize` argument with
#>   non-Gaussian models.
```

<sup>Created on 2025-03-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Do we have a PR approval @DominiqueMakowski ? Merge requires reviewer approval

DominiqueMakowski (approved):
LGTM! @strengejacke one last double-confirm from you?

strengejacke:
I'll rerun tests to update snapshots. I changed some names, to conform with our typical easystats conventions. I think `"bootES"` is not needed as type-name, `"boot"` seems to be enough (or maybe `"bootstrap"`?).

strengejacke:
We have two functions in two files:
`.estimate_contrasts_effecsize` and `.estimate_contrasts_effectsize`. Which is the correct one, the one with the typo, or the other one?

R/estimate_contrasts_effectsize.R
```diff
@@ -0,0 +1,92 @@
+.estimate_contrasts_effecsize <- function(model,
```
strengejacke, line 1: Duplicated function with typo, slightly differ, which one can be removed?

strengejacke (changes requested):

DominiqueMakowski:
Also, about `marginal_d`, just to confirm that it follows our conventions and is line with the other D type names?

strengejacke:
I think we should wait a bit with this PR. If you look at the outputs (see snapshots in files), column names are completely inconsistent across effect sizes and do not comply with our conventions. We should probably fix that first before merging.

rempsyc:
Note: I had changed `attr` to `out_attr` and `by` to `out_by` in `print.R` because using `attr` and `by` triggered lints because they are existing functions. Should we use then `.attr` and `.by` instead? (that's only for internal objects, not argument names)

Also, what is the reason Daniel for putting long code lines on a single line with `# nolint` instead of splitting over several lines? Is it because it creates problems in the snapshots printing?

Finally, could you point me to a good place for our conventions so I could spend some time matching them?

strengejacke:
I think names are ok, must be in _insight_'s `format_table()` to do the final formatting.

lints must be ignored in this package, we use function names as variable names everywhere throughout the package, we can't change every instance. It doesn't make much sense to modify just one file ;-)

> Also, what is the reason Daniel for putting long code lines on a single line with # nolint instead of splitting over several lines? Is it because it creates problems in the snapshots printing?

I think it was laziness, because most tests were already written by other people. But the snapshots were different, and I couldn't figure out the errors in the tests, that's why I overwrote that test file in this PR with the working tests from the main branch, despite overwriting all the code-style edits done here.

Code style conventions: https://easystats.github.io/easystats/articles/conventions.html

I think we can merge this PR-

rempsyc:
Great, thanks a lot @strengejacke! ;)

PR #280, opened 2024-12-16
Draft marginalcontrasts

strengejacke:

PR #287, opened 2025-01-02
Predict auxiliary parameters: estimate_parameters() / "what" argument?

### LINKED ISSUE #286: Predict auxiliary parameters: estimate_parameters() / "what" argument?
I think there's a blindspot in our functionalities.

I am trying to do the following: to predict the value of some auxiliary parameter (in the example below, **sigma**), at a datagrid (in the example below, for each **random factor level**, the expected value of **sigma** at the different levels of *Condition*

``` r
library(brms)

df <- data.frame(
  Condition = rep(c("A", "B", "C"), each=30),
  Participant = rep_len(paste0("S", 1:10), 90),
  Response = rnorm(90)
)

model <- bf(Response ~ Condition + (Condition | Participant),
            sigma ~ Condition + (Condition | Participant),
            family = exgaussian()) |>
  brm(data = df,
    iter = 2000,
    chains = 4,
    cores = 4,
    backend="cmdstanr")


modelbased::estimate_prediction(model, include_random=TRUE, by = c("Condition", "Participant"))
#> Model-based Prediction
#> 
#> Condition | Participant | Predicted |   SE |        95% CI
#> ----------------------------------------------------------
#> A         |          S1 |      0.14 | 1.06 | [-1.84, 2.36]
#> B         |          S1 |      0.08 | 1.54 | [-2.88, 3.18]
#> C         |          S1 |      0.44 | 1.12 | [-1.76, 2.76]
#> A         |          S2 |      0.16 | 0.97 | [-1.67, 2.19]
#> B         |          S2 |      0.09 | 1.12 | [-2.03, 2.44]
#> C         |          S2 |     -0.09 | 1.24 | [-2.38, 2.54]
#> A         |          S3 |      0.28 | 1.01 | [-1.67, 2.38]
#> B         |          S3 |      0.12 | 1.12 | [-2.03, 2.49]
#> C         |          S3 |     -0.04 | 1.03 | [-1.90, 2.17]
#> A         |          S4 |      0.21 | 0.94 | [-1.64, 2.22]
#> B         |          S4 |      0.11 | 1.40 | [-2.69, 2.96]
#> C         |          S4 |      0.34 | 0.95 | [-1.46, 2.42]
#> A         |          S5 |      0.20 | 0.91 | [-1.46, 2.10]
#> B         |          S5 |      0.05 | 1.12 | [-2.04, 2.28]
#> C         |          S5 |      0.18 | 0.90 | [-1.57, 2.07]
#> A         |          S6 |      0.24 | 0.91 | [-1.43, 2.21]
#> B         |          S6 |      0.11 | 0.95 | [-1.71, 2.10]
#> C         |          S6 |     -0.40 | 1.00 | [-2.16, 1.80]
#> A         |          S7 |      0.24 | 0.91 | [-1.38, 2.22]
#> B         |          S7 |      0.24 | 1.03 | [-1.73, 2.46]
#> C         |          S7 |      1.04 | 1.14 | [-1.43, 3.18]
#> A         |          S8 |      0.26 | 0.95 | [-1.48, 2.24]
#> B         |          S8 |      0.15 | 1.10 | [-2.02, 2.47]
#> C         |          S8 |     -0.02 | 1.14 | [-2.15, 2.36]
#> A         |          S9 |      0.19 | 1.09 | [-1.88, 2.40]
#> B         |          S9 |      0.08 | 1.34 | [-2.54, 2.74]
#> C         |          S9 |     -0.06 | 1.16 | [-2.21, 2.28]
#> A         |         S10 |      0.26 | 0.87 | [-1.30, 2.17]
#> B         |         S10 |      0.24 | 0.93 | [-1.54, 2.24]
#> C         |         S10 |      0.56 | 0.92 | [-1.17, 2.48]
#> 
#> Variable predicted: Response
#> Predictors modulated: Condition, Participant
```

<sup>Created on 2025-01-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

In the above, we are predicting the outcome (and its median), but how do we predict the parameters *themselves*?

1) how to achieve that using marginaleffects/marginalmeans? does ggeffects have a solution?
2) Should we add an argument to existing functions to be able to predict not the *outcome* but the value of a given parameter?

strengejacke:
Fixes #286


DominiqueMakowski:
But isn't it overkill to add a function for that?  Isn't it better to add a dpar-like argument to estimate_predictions?

bwiernik:
Where do we currently have prediction of a group-level random slope (if anywhere)?

DominiqueMakowski:
estimate_grouplevel()? https://github.com/easystats/modelbased/blob/main/R/estimate_grouplevel.R

strengejacke:
> But isn't it overkill to add a function for that?

I have no strong opinion here, I thought it was what you had in mind (based on your issue title). We can just do nothing, because `dpar` is passed via dots anyway.

DominiqueMakowski:
- Adding a whole function might increase maintenance burden for quite a small feature
- On the other hand, relying on `dpar` might be a bit too invisible 

- 1) either we document `dpar` in estimate_prediction's kwargs (specifying that it only works when backend is marginaleffects ((which we should discuss whether to make the default)))
- 2) we add an argument to get_predicted 

I'm leaning for option 2, but not sure about the best name. We have `trend`, `contrast` in estimate_slope and estimate_contrasts. So maybe `predict="outcome"` for `estimate_predictions`?

Thoughts @bwiernik @mattansb ?

mattansb:
dpar is documented in [`brms` support functions for `emmeans`](https://paulbuerkner.com/brms/reference/emmeans-brms-helpers.html) as well as in the [Bayesian docs in `marginaleffects`](https://marginaleffects.com/bonus/brms.html#distributional-parameters) so I think it would be fine here as well - better to stick to the convention.

DominiqueMakowski:
> How about we merge this function and grouplevel into one? 

estimate_grouplevel() stays the same, it is for extraction directly the random effects. Here it's mostly about estimating *predicted* parameter values (including over random levels) at a datagrid (so not directly the "effects" per se).

> I think it would be fine here as well - better to stick to the convention

Not a *big* fan of `dpar` (I suppose it stands for Dependent PARameter?) but yeah if it's well established no probs. Then the question is whether to expose it or simply document it in the kwargs

mattansb:
It stands for Distributional PARrameter (:

strengejacke:
> Isn't it better to add a dpar-like argument to estimate_predictions?

Yes, we could add it to `estimate_predictions()` and `estimate_means()` - this would be my preferred choice. And what about `estimate_relation()`? Or do we always need to calculate _prediction intervals_?

> we add an argument to get_predicted

I think it's better to have an argument here in _modelbased_, and then pass `dpar` to `get_predicted()` (or marginaleffects or emmeans via `estimate_means()`). We need the same for `estimate_contrasts()` (and probably `estimate_slopes()`?).

strengejacke:
> 1. which we should discuss whether to make the default

For contrasts, we can use the `by` argument to "filter" contrasts by a predictor's levels. To do so with the marginaleffects backend, we need to reshape the parameter-column, see also https://github.com/easystats/modelbased/pull/280#issuecomment-2548136064

This is quite a tedious work to implement, I guess. But this is an important feature, else the returned output can become _very_ long.

bwiernik:
> Yes, we could add it to estimate_predictions() and estimate_means() - this would be my preferred choice (instead of just documenting dpar). And what about estimate_relation()? Or do we always need to calculate prediction intervals?

This is why I think it would be better to put the parameter estimation in a separate function. Invoking dpar inherently means we are asking for confidence intervals on the distributional parameter. It's a different estimand from relation/expectation/prediction (which are all about the outcome variable).

DominiqueMakowski:
>  It's a different estimand from relation/expectation/prediction (which are all about the outcome variable).

Although I do agree that it's an important conceptual difference, I think from a user-perspective it works to get these using the other functions. Like `estimate_means(predict="sigma")` it still makes "sense" to do that to obtain the "average/mean" of sigma at the different levels

I guess what I'm saying is that keeping with the currently existing API and introducing a new argument instead of a new function would retain its intuitiveness (but yes, we should then document that if `predict` is not `"default"` then CIs etc are about the distributional parameters.

strengejacke:
We don't have a `predict` argument right now. It's used internally, depending on the function that is called. Introducing a `predict` argument would mean we don't need `estimate_expected()`, `estimate_prediction()` etc. anymore. That's why I would a) choose a different name for that argument, b) do nothing and document that `dpar` is passed to `insight::get_predicted()` (though that's the least user-friendly / visible option), or c) introduce a new function (like suggested in this PR).

DominiqueMakowski:
Mmh would it conceptually make sense to add the option directly to `insight::get_predicted` for the [`predict`](https://github.com/easystats/insight/blob/e94ec60f0205ebfdbeca117d621ea1fcdca7c69b/R/get_predicted.R#L29-L47) argument to be able to take values like `sigma` etc.?

Because this argument already changes both the transformation + the nature of CIs etc. so it would make sense to use that to predict other distributional parameters?

strengejacke:
> Mmh would it conceptually make sense to add the option directly to insight::get_predicted for the argument to be able to take values like sigma etc.?

But then we need to include _all_ possible options for `dpar`. Are those documented somewhere? Because we check the `predict` argument for valid options: https://github.com/easystats/insight/blob/e94ec60f0205ebfdbeca117d621ea1fcdca7c69b/R/get_predicted_args.R#L79-L117

strengejacke:
Ok, added to _insight_: https://github.com/easystats/insight/pull/988

DominiqueMakowski:
Now that https://github.com/easystats/insight/pull/988, it should work for estimate_prediction etc., we just need to add & dispatch `predict` to marginaleffects' dpar in `estimate_means()` right?

strengejacke:
Yes, we need to decide a) if we want an additional argument (and if so, which name), or, like Brenton suggested and implemented in this PR, b) have an extra function.

DominiqueMakowski:
since it's now handled by `predict` in `get_prediction`, I'd stick with that. I don't t think we need a new function since this argument could exist in all existing functions (with the caveat that it would probably render the distinction `estimate_relation` and `estimate_expectation` / `estimate_expectation` / `estimate_predicted` less relevant (aside from their difference in the default newdata)

strengejacke:
One counter argument would be that `estimate_relation()`, `estimate_expectation()`, `estimate_expectation()` and `estimate_predicted()` are all the same anyway, they just differ in the default value for their `data` argument, and internally just use different options for the `predict` argument. So why not adding another function (like `estimate_parameters()`) in order to keep the "relevance" of the existing functions?

I have no strong preference here, but if the all the existing function only differ in their default values for arguments, why not just have one integrated function?

On the other hand, to make it work with marginaleffects / emmeans, we need to pass `predict` to `estimate_means()` anyway - or we also add a "backend" argument to `estimate_parameter()`?

hm...

DominiqueMakowski:
estimate_relation(), estimate_expectation() are indeed "aliases" of specific parameters of `estimate_predicted()`, but they are still relevant when one is interested in predicting other dpars.

Basically, predicting other dpars is not orthogonal to the goal of `estimate_relation()` / `estimate_predicted()` etc, you could be interested in predicting other dpars for the original data, for a datagrid, for marginal means etc. etc. 

Hence conceptually I think varying what dpar is predicted fits more as an argument to existing functions rather than a separate function

strengejacke:
That sounds reasonable.

strengejacke:
> Basically, predicting other dpars is not orthogonal to the goal of estimate_relation() / estimate_predicted() etc, you could be interested in predicting other dpars for the original data, for a datagrid, for marginal means etc. etc.

But... the `estimate_()` functions set the value of `predict` (`"expectation"`, `"link"` or `"prediction"`). How would we combine this with `predict = "sigma"`?

DominiqueMakowski:
- estimate_prediction(predict="sigma") & estimate_expectation(predict="sigma")

```r
.estimate_predicted(
    model,
    data = data,
    ci = ci,
    keep_iterations = keep_iterations,
    predict = predict,
    ...
  )
```

- estimate_link(predict="sigma") & estimate_relation(predict="sigma"): (= data is grid)

```r
.estimate_predicted(
    model,
    data = "grid",
    ci = ci,
    keep_iterations = keep_iterations,
    predict = predict,
    ...
  )
```

DominiqueMakowski:
as far as I understand, setting `predict=` to some other dpars renders pairs of existing functions similar and only differering via their default data arg

strengejacke:
Wait, with `posterior_epred()`, `posterior_prediction()` or `predict.brmsfit()`, you can predict the expected values only, to it doesn't make sense to call `estimate_link()` or `estimate_expectation()` etc.?

``` r
m <- insight::download_model("brms_sigma_2")

set.seed(123)
modelbased::estimate_relation(m, include_random = TRUE)
#> Loading required package: rstan
#> Loading required package: StanHeaders
#> 
#> rstan version 2.32.6 (Stan version 2.32.2)
#> For execution on a local, multicore CPU with excess RAM we recommend calling
#> options(mc.cores = parallel::detectCores()).
#> To avoid recompilation of unchanged Stan programs, we recommend calling
#> rstan_options(auto_write = TRUE)
#> For within-chain threading using `reduce_sum()` or `map_rect()` Stan functions,
#> change `threads_per_chain` option:
#> rstan_options(threads_per_chain = 1)
#> Do not specify '-march=native' in 'LOCAL_CPPFLAGS' or a Makevars file
#> Model-based Expectation
#> 
#> Condition | Participant | Predicted |   SE |        95% CI
#> ----------------------------------------------------------
#> A         |          S1 |      0.01 | 0.26 | [-0.49, 0.54]
#> B         |          S1 |      0.20 | 0.32 | [-0.44, 0.82]
#> C         |          S1 |      0.19 | 0.26 | [-0.31, 0.72]
#> A         |          S2 |      0.02 | 0.25 | [-0.47, 0.51]
#> B         |          S2 |     -0.04 | 0.31 | [-0.68, 0.54]
#> C         |          S2 | -9.36e-03 | 0.33 | [-0.70, 0.59]
#> A         |          S3 |     -0.10 | 0.25 | [-0.64, 0.35]
#> B         |          S3 |      0.07 | 0.29 | [-0.50, 0.63]
#> C         |          S3 |     -0.03 | 0.27 | [-0.57, 0.48]
#> A         |          S4 |     -0.03 | 0.23 | [-0.49, 0.44]
#> B         |          S4 |      0.49 | 0.32 | [-0.09, 1.14]
#> C         |          S4 | -2.09e-03 | 0.27 | [-0.51, 0.54]
#> A         |          S5 |     -0.02 | 0.21 | [-0.43, 0.40]
#> B         |          S5 |      0.26 | 0.24 | [-0.23, 0.75]
#> C         |          S5 |      0.02 | 0.28 | [-0.53, 0.59]
#> 
#> Variable predicted: Response
#> Predictors modulated: Condition, Participant
set.seed(123)
modelbased::estimate_link(m, include_random = TRUE)
#> Model-based Link
#> 
#> Condition | Participant | Predicted |   SE |        95% CI
#> ----------------------------------------------------------
#> A         |          S1 |      0.01 | 0.26 | [-0.49, 0.54]
#> B         |          S1 |      0.20 | 0.32 | [-0.44, 0.82]
#> C         |          S1 |      0.19 | 0.26 | [-0.31, 0.72]
#> A         |          S2 |      0.02 | 0.25 | [-0.47, 0.51]
#> B         |          S2 |     -0.04 | 0.31 | [-0.68, 0.54]
#> C         |          S2 | -9.36e-03 | 0.33 | [-0.70, 0.59]
#> A         |          S3 |     -0.10 | 0.25 | [-0.64, 0.35]
#> B         |          S3 |      0.07 | 0.29 | [-0.50, 0.63]
#> C         |          S3 |     -0.03 | 0.27 | [-0.57, 0.48]
#> A         |          S4 |     -0.03 | 0.23 | [-0.49, 0.44]
#> B         |          S4 |      0.49 | 0.32 | [-0.09, 1.14]
#> C         |          S4 | -2.09e-03 | 0.27 | [-0.51, 0.54]
#> A         |          S5 |     -0.02 | 0.21 | [-0.43, 0.40]
#> B         |          S5 |      0.26 | 0.24 | [-0.23, 0.75]
#> C         |          S5 |      0.02 | 0.28 | [-0.53, 0.59]
#> 
#> Variable predicted: Response
#> Predictors modulated: Condition, Participant
```

<sup>Created on 2025-01-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

That means, if we don't need the different functions for _different predicted values of dpar_, we don't need a `predict` argument for the different functions. In this case, it would indeed make more sense to just have a separate function?

@mattansb ?

bwiernik:
This is why I don't think it makes sense to put these into relation/expectation/prediction (setting aside means/slopes/contrasts for now). 

Relation/expectation/prediction are all about estimating predicted outcomes and they differ in whether they return the predicted link function, expected value, or natural-scale predicted value and their associated intervals. Analogous to brms posterior_linpred,_epred, and _prediction. 

If I want predicted sigma values, I'm not requesting any of those three--there isn't a distinction between expectation and prediction. It feels very hacky to me in brms that the dpar argument overrides other arguments in the posterior_ functions. I would rather keep the different estimands separate.

DominiqueMakowski:
It is hacky, but if we make clear that if `predict="sigma"` then the distinction between expectation and prediction disappears I think it's good enough

I think that's a case of conceptual correctness vs. UX, and I'm on the side of the prioritizing the user API here so I'm fine with the "hack"

strengejacke:
>  I'm on the side of the prioritizing the user API here

But isn't a new argument, if set to (e.g.) `"sigma"`, which returns the _same_ estimates for any of the four functions (expectation/link/relation/prediction) is _more_ confusing than a separate function?

But I see the point, we can't be 100% consistent here - if we _also_ want to predict sigma using emmeans or marginaleffects, we a) _need_ to call the `estimate_means()` function (i.e. we need a new argument `predict` or whichever name) or b) have a `backend` argument in `estimate_parameter()`, which internally calls emmeans or marginaleffects (or insight::get_predicted, if not set).

And then, as Brenton mentioned, we still have `estimate_slopes()` (though I'm not sure whether you can estimate a "trend" for sigma? maybe...)

I have no 100% opinion right now, but Dom, you're the maintainer, the final decision is up to you.

DominiqueMakowski:
>  same estimates for any of the four functions (expectation/link/relation/prediction) is more confusing than a separate function?

I don't think so. The distinction between expectation/prediction/link is already subtle enough and gets tricky regardless with GLMs

> if we also want to predict sigma using emmeans or marginaleffects, we a) need to call the estimate_means() function (i.e. we need a new argument predict or whichever name)

Yes I would be happy to also have the same `predict` argument in `estimate_means()/contrasts` (albeit here in the docs we can focus on its use as `dpar`

--- 

Basically, for the functions related to `insight::get_predicted()`, the `predict` argument was pretty much hidden anyway for users (and was mainly modulated by the different subfunctions). Now, we can expose it a bit more and explain its role in getting other dpars. 

For functions related to marginalmeans/emmeans, we add the `predict` argument, focusing on its use for getting other dpars.

And everything is nice and consistent and they lived happily ever after

R/estimate_predicted.R
```diff
@@ -143,9 +143,7 @@
 #' types). The `by` argument will be used to create a data grid via
 #' `insight::get_datagrid()`, which will then be used as `data` argument. Thus,
 #' you cannot specify both `data` and `by` but only of these two arguments.
-#' @param parameter Name of the auxiliary parameter to estimate. Defaults to
-#' `"sigma"`, but can be any other value that is usually accepted by the `dpar`
-#' argument.
+#' @param predict to do...
```
DominiqueMakowski, line 146:
```suggestion
#' @param predict This parameter controls what is predicted. In most cases, it is changed automatically when using the different functions such as `estimate_expectation()`,  `estimate_prediction()` or `estimate_relation()`. The only time you might be interested in manually changing it is to estimate other distributional parameters (called "dpar" in other packages) - for instance when using complex formulae in `brms` models. The `predict` argument can take the value of the parameter you want to estimate, for instance `"sigma"`, `"kappa"`, etc. This argument gets passed to [insight::get_predicted()].
```

strengejacke:
Here we go:

``` r
library(brms)
library(ggeffects)
library(easystats)
m <- insight::download_model("brms_sigma_2")

# ggeffects
out <- predict_response(
  m,
  c("Condition", "Participant"),
  type = "random",
  dpar = "sigma"
)
print(out, collapse_tables = TRUE)
#> # Predicted values of Response
#> 
#> Condition | Participant | Predicted |     95% CI
#> ------------------------------------------------
#> A         |          S1 |      0.99 | 0.45, 1.79
#> B         |             |      0.79 | 0.26, 1.58
#> C         |             |      0.55 | 0.09, 1.29
#> A         |          S2 |      0.73 | 0.25, 1.33
#> B         |             |      0.63 | 0.30, 1.22
#> C         |             |      0.96 | 0.57, 1.95
#> A         |          S3 |      0.92 | 0.45, 1.61
#> B         |             |      0.72 | 0.32, 1.40
#> C         |             |      0.68 | 0.25, 1.36
#> A         |          S4 |      0.69 | 0.21, 1.27
#> B         |             |      0.61 | 0.11, 1.32
#> C         |             |      0.63 | 0.09, 1.34
#> A         |          S5 |      0.59 | 0.07, 1.19
#> B         |             |      0.44 | 0.01, 1.04
#> C         |             |      0.73 | 0.08, 1.69

# marginaleffects
marginaleffects::predictions(
  m,
  newdata = data_grid(m, c("Condition", "Participant")),
  by = c("Condition", "Participant"),
  dpar = "sigma"
)
#> 
#>  Condition Participant Estimate  2.5 % 97.5 %
#>          A          S1    0.987 0.4456   1.79
#>          A          S2    0.733 0.2520   1.33
#>          A          S3    0.916 0.4518   1.61
#>          A          S4    0.693 0.2064   1.27
#>          A          S5    0.588 0.0652   1.19
#>          B          S1    0.792 0.2634   1.58
#>          B          S2    0.628 0.3011   1.22
#>          B          S3    0.715 0.3223   1.40
#>          B          S4    0.613 0.1113   1.32
#>          B          S5    0.441 0.0147   1.04
#>          C          S1    0.547 0.0927   1.29
#>          C          S2    0.957 0.5672   1.95
#>          C          S3    0.684 0.2464   1.36
#>          C          S4    0.635 0.0917   1.34
#>          C          S5    0.728 0.0833   1.69
#> 
#> Type:  response 
#> Columns: rowid, Condition, Participant, estimate, conf.low, conf.high, Response

# marginaleffects
modelbased::estimate_means(
  m,
  by = c("Condition", "Participant"),
  predict = "sigma",
  backend = "marginaleffects"
)
#> Estimated Marginal Means
#> 
#> Condition | Participant | Mean |       95% CI
#> ---------------------------------------------
#> A         |          S1 | 0.99 | [0.45, 1.79]
#> A         |          S2 | 0.73 | [0.25, 1.33]
#> A         |          S3 | 0.92 | [0.45, 1.61]
#> A         |          S4 | 0.69 | [0.21, 1.27]
#> A         |          S5 | 0.59 | [0.07, 1.19]
#> B         |          S1 | 0.79 | [0.26, 1.58]
#> B         |          S2 | 0.63 | [0.30, 1.22]
#> B         |          S3 | 0.72 | [0.32, 1.40]
#> B         |          S4 | 0.61 | [0.11, 1.32]
#> B         |          S5 | 0.44 | [0.01, 1.04]
#> C         |          S1 | 0.55 | [0.09, 1.29]
#> C         |          S2 | 0.96 | [0.57, 1.95]
#> C         |          S3 | 0.68 | [0.25, 1.36]
#> C         |          S4 | 0.63 | [0.09, 1.34]
#> C         |          S5 | 0.73 | [0.08, 1.69]
#> 
#> Marginal means estimated at Condition

# backend emmeans fails
modelbased::estimate_means(
  m,
  by = c("Condition", "Participant"),
  predict = "sigma"
)
#> Error: Couldn't find which variables were selected in `by`. Check spelling and
#>   specification.

# insight::get_predicted
modelbased::estimate_relation(
  m,
  by = c("Condition", "Participant"),
  predict = "sigma",
) |> data_arrange("Condition")
#> Model-based Sigma
#> 
#> Condition | Participant | Predicted |   SE |       95% CI
#> ---------------------------------------------------------
#> A         |          S1 |      1.02 | 0.34 | [0.45, 1.79]
#> A         |          S2 |      0.75 | 0.27 | [0.25, 1.33]
#> A         |          S3 |      0.95 | 0.29 | [0.45, 1.61]
#> A         |          S4 |      0.70 | 0.27 | [0.21, 1.27]
#> A         |          S5 |      0.59 | 0.31 | [0.07, 1.19]
#> B         |          S1 |      0.83 | 0.33 | [0.26, 1.58]
#> B         |          S2 |      0.66 | 0.24 | [0.30, 1.22]
#> B         |          S3 |      0.75 | 0.27 | [0.32, 1.40]
#> B         |          S4 |      0.63 | 0.32 | [0.11, 1.32]
#> B         |          S5 |      0.45 | 0.28 | [0.01, 1.04]
#> C         |          S1 |      0.59 | 0.31 | [0.09, 1.29]
#> C         |          S2 |      1.03 | 0.37 | [0.57, 1.95]
#> C         |          S3 |      0.72 | 0.29 | [0.25, 1.36]
#> C         |          S4 |      0.65 | 0.31 | [0.09, 1.34]
#> C         |          S5 |      0.76 | 0.42 | [0.08, 1.69]
#> 
#> Variable predicted: Response
#> Predictors modulated: Condition, Participant

# insight::get_predicted
modelbased::estimate_link(
  m,
  by = c("Condition", "Participant"),
  predict = "sigma"
) |> data_arrange("Condition")
#> Model-based Sigma
#> 
#> Condition | Participant | Predicted |   SE |       95% CI
#> ---------------------------------------------------------
#> A         |          S1 |      1.02 | 0.34 | [0.45, 1.79]
#> A         |          S2 |      0.75 | 0.27 | [0.25, 1.33]
#> A         |          S3 |      0.95 | 0.29 | [0.45, 1.61]
#> A         |          S4 |      0.70 | 0.27 | [0.21, 1.27]
#> A         |          S5 |      0.59 | 0.31 | [0.07, 1.19]
#> B         |          S1 |      0.83 | 0.33 | [0.26, 1.58]
#> B         |          S2 |      0.66 | 0.24 | [0.30, 1.22]
#> B         |          S3 |      0.75 | 0.27 | [0.32, 1.40]
#> B         |          S4 |      0.63 | 0.32 | [0.11, 1.32]
#> B         |          S5 |      0.45 | 0.28 | [0.01, 1.04]
#> C         |          S1 |      0.59 | 0.31 | [0.09, 1.29]
#> C         |          S2 |      1.03 | 0.37 | [0.57, 1.95]
#> C         |          S3 |      0.72 | 0.29 | [0.25, 1.36]
#> C         |          S4 |      0.65 | 0.31 | [0.09, 1.34]
#> C         |          S5 |      0.76 | 0.42 | [0.08, 1.69]
#> 
#> Variable predicted: Response
#> Predictors modulated: Condition, Participant

# brms::posterior_epred
dg <- data_grid(m, c("Condition", "Participant"))
out <- cbind(
  dg,
  data.frame(
    predicted = posterior_epred(
      m,
      newdata = dg,
      dpar = "sigma"
    ) |> colMeans() |> 
      round(2)
  )
)
data_arrange(out, "Condition")
#>    Condition Participant predicted
#> 1          A          S1      1.02
#> 2          A          S2      0.75
#> 3          A          S3      0.95
#> 4          A          S4      0.70
#> 5          A          S5      0.59
#> 6          B          S1      0.83
#> 7          B          S2      0.66
#> 8          B          S3      0.75
#> 9          B          S4      0.63
#> 10         B          S5      0.45
#> 11         C          S1      0.59
#> 12         C          S2      1.03
#> 13         C          S3      0.72
#> 14         C          S4      0.65
#> 15         C          S5      0.76
```

<sup>Created on 2025-01-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Not sure why `ggeffects::predcict_response()` is similar to _marginaleffects_, and not `insight::get_predicted()`, because it internally also calls `posterior_epred()`... But anyway, seems to work.

strengejacke:
In `estimate_means()`, when `predict` is not a dpar, we have two arguments with (almost) the same purpose: `transform` and `predict`.

DominiqueMakowski:
> In estimate_means()

We could remove `transform` in favour of predict, either it would be `response` (by default) or `link` for the untransformed, or dpar?

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Type: Package
 Package: modelbased
 Title: Estimation of Model-Based Predictions, Contrasts and Means
-Version: 0.8.9.3
+Version: 0.8.9.4
```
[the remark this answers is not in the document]
strengejacke: to make `easystats::install_latest()` work properly, I would set the final version short before submission. Else, for every PR, increase the 4th digit.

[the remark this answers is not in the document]
DominiqueMakowski: alright!
BUT STILL I can feel we're getting very close (thanks to you 😁)

strengejacke:
WDYT, should the column be named "Mean" (formerly) or like the dpar, in this case "Sigma"?

``` r
m <- insight::download_model("brms_sigma_2")
modelbased::estimate_means(
  m,
  by = c("Condition", "Participant"),
  predict = "sigma",
  backend = "marginaleffects"
)
#> Estimated Marginal Means
#> 
#> Condition | Participant | Sigma |       95% CI
#> ----------------------------------------------
#> A         |          S1 |  0.99 | [0.45, 1.79]
#> A         |          S2 |  0.73 | [0.25, 1.33]
#> A         |          S3 |  0.92 | [0.45, 1.61]
#> A         |          S4 |  0.69 | [0.21, 1.27]
#> A         |          S5 |  0.59 | [0.07, 1.19]
#> B         |          S1 |  0.79 | [0.26, 1.58]
#> B         |          S2 |  0.63 | [0.30, 1.22]
#> B         |          S3 |  0.72 | [0.32, 1.40]
#> B         |          S4 |  0.61 | [0.11, 1.32]
#> B         |          S5 |  0.44 | [0.01, 1.04]
#> C         |          S1 |  0.55 | [0.09, 1.29]
#> C         |          S2 |  0.96 | [0.57, 1.95]
#> C         |          S3 |  0.68 | [0.25, 1.36]
#> C         |          S4 |  0.63 | [0.09, 1.34]
#> C         |          S5 |  0.73 | [0.08, 1.69]
#> 
#> Marginal means estimated at Condition
```

<sup>Created on 2025-01-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
We have to set the default for `predict` to `NULL` in those methods that rely on emmeans or marginaleffects, because we need different default values for each package, and for marginaleffects also based on the model type. The correct `predict` option is defined in `.get_emmeans_type_argument()` and `.get_type_argument()`

mattansb:
The column name should be a mix of the parameter's name and the point estimate we're using - "nu_median", "sigma_mean", "phi_MAP", etc...

strengejacke:
> The column name should be a mix of the parameter's name and the point estimate we're using - "nu_median", "sigma_mean", "phi_MAP", etc...

Can we set the centrality somewhere in _marginaleffects_? Or do we have any information on whether Median, Mean etc. are returned?

strengejacke:
Just realized that adding a new function would have been less effort. :-)
Should be working now, but we still have to polish some things (docs, add more tests, make sure we have proper outputs/column names)...

When tests pass, I would merge this PR and then open a new one for the remaining details.

strengejacke:
btw, @DominiqueMakowski, should we remove the aliases `model_*()`? Else, we have three different names for one thing:

`estimate_means()`
`get_marginalmeans()`
`model_marginalmeans()`

mattansb:
> Can we set the centrality somewhere in marginaleffects? Or do we have any information on whether Median, Mean etc. are returned?

Not sure about `marginaleffects`, but `bayestestR` fully supports outputs now, so we can control that ourselves...

strengejacke:
So I would pass the result from `avg_predictions()` to `describe_posterior()`?

DominiqueMakowski:
> WDYT, should the column be named "Mean" (formerly) or like the dpar, in this case "Sigma"?

"Sigma"

Not sure about adding the point estimate, I don't like it for aesthetic reasons and we don't do it here (i.E., we still write "Mean" fro marginal means when the point estimate is median)

DominiqueMakowski:
> should we remove the aliases model_*()?

The `get_*` ones do something else, they are the wrappers around marginaleffects/emmeans that the estimate_* functions use. But yes, we should remove the `model_*` for sure

mattansb:
> So I would pass the result from avg_predictions() to describe_posterior()?

Yes, I think so.

R/clean_names.R
```diff
@@ -53,3 +55,8 @@
   means$Parameter <- NULL
   means
 }
+
+
+.capitalize <- function(x) {
```
DominiqueMakowski, line 60: [`tools::toTitleCase()`](https://stackoverflow.com/a/38527665/4198688)?

strengejacke: I knew there was a function, but could not remember... `toTitleCase()` does a bit more, but for a single word, it doesn't matter and has the same result as `.capitalize()`.

strengejacke:
@DominiqueMakowski I think for version 1.0.0, most/all of these should be fixed, if possible. I see most issues with plotting, would be nice if we can fix plotting, as this seems to be an essential feature to me:
https://github.com/easystats/modelbased/issues?q=is%3Aopen+is%3Aissue+milestone%3A1.0.0

And we should definitely fix the printing for contrasts using marginaleffects, else we cannot filter using `by`:
https://github.com/easystats/modelbased/blob/26a4620863548860972017ed739a1097c472027e/R/estimate_contrasts.R#L188-L236

PR #290, opened 2025-01-07
[Fix] estimate_contrasts(..., predict="response") #215

### LINKED ISSUE #215: estimate_contrasts with transform = "response" for a logistic model does not give estimates in terms of probabilities
When doing estimate_contrasts with transform = "response" for a logistic model the estimates are not in terms of probabilities like the help page said:

"Thus for a logistic model, "none" will give estimations expressed in log-odds (probabilities on logit scale) and "response" in terms of probabilities."

```r
g <- glm(vs ~ factor(am), data = mtcars, family = binomial())

estimate_contrasts(g, contrast = "am", transform = "response")

Marginal Contrasts Analysis

Level1 | Level2 | Difference |        95% CI |   SE |  df |     z |     p
-------------------------------------------------------------------------
am0    |    am1 |      -0.69 | [-2.13, 0.74] | 0.73 | Inf | -0.95 | 0.344

Marginal contrasts estimated at am
p-value adjustment method: Holm (1979)
```

DominiqueMakowski:
Fixes #215

Previously, for a logistic model, contrasts on the response scale returned Odds ratio (category A / B). Following  #215, we can regrid that and get actual contrasts on the response scale

``` r
  # GLM - binomial
  dat <- iris
  dat$y <- as.factor(ifelse(dat$Sepal.Width > 3, "A", "B"))
  dat <<- dat
  model <- glm(y ~ Species, family = "binomial", data = dat)
    
  # Default
  modelbased::estimate_contrasts(model)
#> No variable was specified for contrast estimation. Selecting `contrast =
#>   "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |         95% CI |   SE |  df |     z |      p
#> -----------------------------------------------------------------------------------
#> setosa     | versicolor |      -3.32 | [-4.62, -2.01] | 0.55 | Inf | -6.08 | < .001
#> setosa     |  virginica |      -2.32 | [-3.49, -1.15] | 0.49 | Inf | -4.76 | < .001
#> versicolor |  virginica |       0.99 | [-0.17,  2.16] | 0.49 | Inf |  2.04 | 0.041 
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)
  
  # New 
  modelbased::estimate_contrasts(model, predict = "response")
#> No variable was specified for contrast estimation. Selecting `contrast =
#>   "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |         95% CI |   SE |  df |     z |      p
#> -----------------------------------------------------------------------------------
#> setosa     | versicolor |      -0.68 | [-0.86, -0.50] | 0.07 | Inf | -9.27 | < .001
#> setosa     |  virginica |      -0.50 | [-0.70, -0.30] | 0.08 | Inf | -5.90 | < .001
#> versicolor |  virginica |       0.18 | [-0.02,  0.38] | 0.08 | Inf |  2.12 | 0.034 
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)
```

<sup>Created on 2025-01-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

But does it make sense? @bwiernik what should be the correct output here?

- [x] Currently setting `predict="response"` doesn't change anything for `marginaleffects` backend


bwiernik:
I think defaulting to contrasts on the response scale is correct. Changing marignaleffects `type` to "link" will return difference in log odds, which we can exponentiate to get odds ratios. We should have that as an option, but should default to differences on the natural scale.

strengejacke:
What are the predicted values on the response scale? The contrasts should be in percentage points between predicted probabilities

DominiqueMakowski:
Just to clarify @bwiernik because there were are 2 things in my mind:

- So the output above seems correct (and we should run [regrid()](https://github.com/easystats/modelbased/pull/290/files#diff-8fa8d68cd5fdb81925415fbaccf8ac670c7de96e894aa542810e1b433b938bceR58) whenever `predict="response"`
- You think `predict="response"` by default? I'm a bit hesitant about that, I always thought it's best to avoid using and applying stats after transformation if possible

DominiqueMakowski:
> What are the predicted values on the response scale? The contrasts should be in percentage points between predicted probabilities

I think that's what they are

bwiernik:
> You think predict="response" by default? I'm a bit hesitant about that, I always thought it's best to avoid using and applying stats after transformation if possible

The most natural method for comparison is the response scale in my opinion. Differences in percentages points are the most interpretable metric and should be the default. marignaleffects's delta method intervals do a good job of estimating the uncertainty in this metric. There are some reasonable alternative views for why (log) odds ratios should be preferred for contrasts, so we should make them an option and document how to get them, but I strongly think response scale differences are the most appropriate default.

R/estimate_contrasts.R
```diff
@@ -77,7 +77,7 @@
 estimate_contrasts <- function(model,
                                contrast = NULL,
                                by = NULL,
-                               predict = NULL,
+                               predict = "response",
```
strengejacke, line 80: Please don't change the option for predict here. Instead, change it inside `get_emmeans_type_argument()`.

strengejacke (changes requested):

R/estimate_contrasts.R
DominiqueMakowski: why? if so it doens't work for marginaleffects backend

DominiqueMakowski: and isn't it clearer when the default argument is in the function signature? I'm not a big fan of defaults that are hard to infer from the docstrings

strengejacke: True, but marginaleffects should not default to "response", but to the first valid type from the type-dictionary (which is handled in `.get_marginaleffects_type_argument()`). This is one of the marginaleffects quirks, see this example (and differing CIs):

``` r
# GLM - binomial
dat <- iris
dat$y <- as.factor(ifelse(dat$Sepal.Width > 3, "A", "B"))
model <- glm(y ~ Species, family = "binomial", data = dat)

marginaleffects::avg_predictions(model, by = "Species", type = "response")
#> 
#>     Species Estimate Std. Error     z Pr(>|z|)     S  2.5 % 97.5 %
#>  setosa         0.16     0.0518  3.09  0.00203   8.9 0.0584  0.262
#>  versicolor     0.84     0.0518 16.20  < 0.001 193.7 0.7384  0.942
#>  virginica      0.66     0.0670  9.85  < 0.001  73.7 0.5287  0.791
#> 
#> Type:  response 
#> Columns: rowid, Species, estimate, std.error, statistic, p.value, s.value, conf.low, conf.high, rowid_dedup
marginaleffects::avg_predictions(model, by = "Species", type = "invlink(link)")
#> 
#>     Species Estimate Pr(>|z|)    S  2.5 % 97.5 %
#>  setosa         0.16   <0.001 15.8 0.0821  0.289
#>  versicolor     0.84   <0.001 15.8 0.7114  0.918
#>  virginica      0.66   0.0263  5.2 0.5195  0.777
#> 
#> Type:  invlink(link) 
#> Columns: rowid, Species, estimate, p.value, s.value, conf.low, conf.high, rowid_dedup
```

<sup>Created on 2025-01-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I added a test, it works correctly for contrasts.

strengejacke: > I'm not a big fan of defaults that are hard to infer from the docstrings

I think for `get_marginalcontrasts()`, it doesn't matter if the default is `response`. But for `get_marginalmeans()`, defaulting to `"response"` returns wrong SEs and CIs in certain situations (see example above), that's why for glm's, marginaleffects defaults to `"invlink(link)"`. So we should _not_ set a default value here, but use the default from the type-dictionary.

Having `predict = NULL` defaults to the most appropriate value now (response for emmeans, and response or invlink for marginaleffects). I also used `NULL`, because the former default of `get_emcontrasts()` was `"link"`, so I added the `.get_em_type()` function, which has the information which function was called and sets the appropriate value for `predict`. Since we now default to `"response"` in `get_emcontrasts()`, we _could_ set `"response"` as default. But this won't work for `estimate_means()` when `backend = "marginaleffects"`, because `"response"` is not always the appropriate option (see example above).

To summarize: I would stick to `predict = NULL`, and we set the best default internally. If the user wants something else, `predict` must be provided.

strengejacke:
Ok, everything seems to work. Question is whether we want `predict = "response"` as default for the emmeans-related-functions, or stick to `NULL` for all? In `estimate_means()`, `predict` _has_ to be `NULL`, in `get_emmeans()`, the default could be `"response"`.

DominiqueMakowski:
> in get_emmeans(), the default could be "response".

I'm fine with either, tbh the  `get_*` functions are mostly dev-oriented wrappers anyway so I don't think it matters much

strengejacke:
Since `predict` is `NULL` by default, the information about the scale is also printed in the footer now.

``` r
dat <- iris
dat$y <- as.factor(ifelse(dat$Sepal.Width > 3, "A", "B"))
dat <<- dat
model <- glm(y ~ Species, family = "binomial", data = dat)
  
modelbased::estimate_means(model)
#> We selected `by = c("Species")`.
#> Estimated Marginal Means
#> 
#> Species    | Probability |   SE |       95% CI
#> ----------------------------------------------
#> setosa     |        0.16 | 0.05 | [0.08, 0.29]
#> versicolor |        0.84 | 0.05 | [0.71, 0.92]
#> virginica  |        0.66 | 0.07 | [0.52, 0.78]
#> 
#> Marginal means estimated at Species
#> Predictions are on the response-scale.

modelbased::estimate_contrasts(model)
#> No variable was specified for contrast estimation. Selecting `contrast =
#>   "Species"`.
#> Marginal Contrasts Analysis
#> 
#> Level1     |     Level2 | Difference |         95% CI |   SE |  df |     z |      p
#> -----------------------------------------------------------------------------------
#> setosa     | versicolor |      -0.68 | [-0.86, -0.50] | 0.07 | Inf | -9.27 | < .001
#> setosa     |  virginica |      -0.50 | [-0.70, -0.30] | 0.08 | Inf | -5.90 | < .001
#> versicolor |  virginica |       0.18 | [-0.02,  0.38] | 0.08 | Inf |  2.12 | 0.034 
#> 
#> Marginal contrasts estimated at Species
#> p-value adjustment method: Holm (1979)
#> Contrasts are on the response-scale.
```

<sup>Created on 2025-01-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

We still can decide whether we want to be more detailed (e.g. _Predictions are shown as probabilities_ ), but for now, this should be ok I think.

bwiernik:
LGTM

Issue #291, opened 2025-01-07
`estimate_slope()` not yet supporting {marginaleffects}


strengejacke:
Currently, it only relies on _emmeans_, we have no `backend` argument.

The `get_marginaleffects()` function needs some more testing, sometimes results between emmeans and marginaleffects differ (I think it was for glmmTMB zero-inflated poisson).

PR #296, opened 2025-01-08
[Feature] marginaleffects backend for estimate_slopes()

DominiqueMakowski:
#291


strengejacke (commented):
> @strengejacke do we need the degrees of freedom?

I think it's better to include them.

DominiqueMakowski:
why not pull them directly from marginaleffects' output?

strengejacke:
We have to pass them to marginaleffects, because its default is _always_ `Inf`.

strengejacke:
Just curious, why don't you just rename `get_marginaleffects()` into `get_marginaltrends()`?

DominiqueMakowski:
Mmh I assumed get_marginaleffects was something else, a basis for get_marginalmeans...

DominiqueMakowski:
Then implementing the backend is just a matter of plugin in into estimate_slopes?

strengejacke:
> Then implementing the backend is just a matter of plugin in into estimate_slopes?

Yes. :-)

You drafted `get_marginaleffects()` three years ago, according to git blame, and looking at the example, the function arguments and now at the code you wrote here in this PR, I think these two are "the same". ;-)

DominiqueMakowski:
and here I was just thinking "Whoever wrote this function really seem to know his shit"

strengejacke:
`get_marginaleffects()` needs some more testing, but the basics are there. It's just that `estimate_slope()` is only for emmeans right now, and has no `backend` argument to call `get_marginaleffects()` (or `get_marginaltrends()`).

DominiqueMakowski:
Should we rename it as `get_marginaltrends()` to not confuse it with the package name?

strengejacke:
Yes, then it's consistent with the `em_` function. And the user-interface is `estimate_slopes()` anyway.

DominiqueMakowski:
oops i pushed to main instead of here

DominiqueMakowski:
> It's just that estimate_slope() is only for emmeans right now, and has no backend argument to call get_marginaleffects() (or get_marginaltrends()).

Since we are both here, should I let you give it a quick go 😁

strengejacke:
> > It's just that estimate_slope() is only for emmeans right now, and has no backend argument to call get_marginaleffects() (or get_marginaltrends()).
> 
> Since we are both here, should I let you give it a quick go 😁

Yeah, I can rename the files :-)

DominiqueMakowski:
image

strengejacke:
I rarely compute slopes/marginaleffects, so we have to test this function in "real life" and fix/improve it whenever we find anything relevant.

DominiqueMakowski:
lgtm, thanks, will merge but will give it  some real life test

strengejacke:
Let me try to fix the failing tests first. Not sure what's going on here

PR #328, opened 2025-01-13
Label axis when data is labelled

### LINKED ISSUE #327: Label axis when data is labelled

strengejacke:
Fixes #327

### DIFF
--- a/R/visualisation_recipe_internal.R
+++ b/R/visualisation_recipe_internal.R
@@ -10,6 +10,10 @@
     group = 1
   )
 
+  # extract information for labels
+  model_data <- try(insight::get_data(attributes(x)$model), silent = TRUE)
+  model_response <- attributes(x)$response
+
   # Main geom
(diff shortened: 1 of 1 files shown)


R/visualisation_recipe_internal.R
```diff
@@ -68,7 +72,35 @@
...
+      colour <- .safe(attr(model_data[[by[[2]]]], "label", exact = TRUE))
+    } else {
+      colour <- NULL
+    }
+    axis_labels <- list(ylab = ylab, xlab = xlab, colour = colour)
+  } else {
+    axis_labels <- NULL
+  }
+
+  list(aes = aes, data = data, axis_labels = axis_labels)
```
DominiqueMakowski, line 103: rather than a separate list, might be tidier to just add `xlab` and `ylab` and `colourlab` inside the first `aes` list

strengejacke: Ok, will do

PR #332, opened 2025-01-13
[paper] writing

DominiqueMakowski:
Let's get this bad boi out.

@mattansb @bwiernik what would you write about the two backends, emmeans and marginaleffects, their main differences etc. 

Do they both rely on the delta method?


DominiqueMakowski:
@strengejacke I think I'll need your help for the marginal section, the different types of marginalization etc. it's quite blurry in my head to be honest ^^

DominiqueMakowski:
@easystats/core-team remaining bits:

- [ ] add details about types of marginalization 
- [ ] Drawbacks/benefits of both backends
- [ ] Polishing

mattansb:
`emmeans` relies on the delta method for non-linear _transformations_ when those are done as part of the estimation.

@DominiqueMakowski where do you want my input, and what should it include? I must admit I have 0 familiarity with `modelbased` - I use `marginaleffects`, `emmeans` and `ggeffects` directly for plotting and estimation, so I have no idea what would be relevant here.

DominiqueMakowski:
>  I must admit I have 0 familiarity 

Your torment over you can make the switch now 😛 

Mostly if you could take a look at the technical details section that talks about the backends

strengejacke:
_modelbased_ currently covers classical `predict()` (non-focal held constant at representative values) or estimates marginal means (non-focal averaged/weighted), the latter using _emmeans_ resp. _marginaleffects_ via `estimate_means()`. `estimate_means()` should return identical results for both _emmeans_ and _marginaleffects_ backends, i.e. currently we support _marginaleffects_ by mimicking the emmeans-approach of marginalzation.

paper/paper.md
```diff
@@ -56,28 +56,181 @@ In line with the `easystats`' *raison d'être*, the `modelbased` package is desi
...
+This is crucial for interpreting how individual predictors influence the response variable in complex models.
+
+- `estimate_means()`: computes **Marginal Means**, i.e., the average predictions for each level of a categorical predictor, averaged across all levels of other predictors.
+- `estimate_contrasts()`: computes **Marginal Contrasts**, i.e., the comparison the marginal means of different levels of a factor to assess differences or effects.
+- `estimate_slopes()`: computes **Marginal Slopes**, i.e., the change in the response variable for a one-unit change in a predictor, holding all other predictors constant. They are essentially the partial derivatives of the response with respect to each predictor, useful for continuous predictors.
+
+The modelbased package simplifies the extraction of these effects, providing a clear interface to understand how different predictors interact with outcomes in various scenarios.
+
+[details about types of marginalization]
```
DominiqueMakowski, line 90: @strengejacke I think here we could mention our 2 main types of marginalization (essentially copypasta our nice docs on that)

paper/paper.md
```diff
@@ -56,28 +56,181 @@ In line with the `easystats`' *raison d'être*, the `modelbased` package is desi
...
 
 The algorithmic heavy lifting is done by its two backend packages, `emmeans` and `marginaleffects`.
 
-Differences? 
+- `emmeans` (REF) Originally, the package was known as `lsmeans`, which stands for "Least-Squares Means". 
+  This term was coined by researchers in the statistical community to describe what are now more commonly referred to as "Estimated Marginal Means" or EMMs, which are essentially predictions averaged over specified levels of factors in the model while holding continuous variables at their means or other specified values.
+  The term "Least-Squares Means" was somewhat misleading as it suggested a method specific to least-squares estimation, hence its renaming to `emmeans` in 2016 to clarify its scope for a wider range of models including generalized linear models, mixed models, and Bayesian models.
+- `marginaleffects` (REF) was more recently introduced and also employs the delta method to approximate variance estimation. It is compatible with a wider range of models and allows for more flexibility in the specification of the marginal effects to be estimated.
+
+[What's the difference / benefits / drawbacks of using one or ther other?]
```
DominiqueMakowski, line 104: @mattansb any other interesting facts to mention here?

paper/paper.md
strengejacke: I can start working on this paper by mid of February, or earlier if it's just minor stuff.

paper/paper.md
strengejacke: Reference for marginaleffects would be https://www.jstatsoft.org/article/view/v111i09

mattansb: Oh there's a lot that I can add. 

I think the biggest difference would be:
- emmeans used a reference grid (that has by default a "complete" combination of all predictors) to generate expected values, by default conditioning on the mean of numeric values and marginalizing over categorical/binary variables, using a linear function of the model's coefficients (and v-cov matrix to get SEs) to give "predictions at the mean" (predictions for an average observation). 
- marginaleffects uses unit level predictions to generate two counterfactual values - the difference of which is then taken (with SEs computed using the delta method), and averaged across all units. By default, the original model frame is used.

Of course, emmeans can also the delta method and can build complex reference grids (that aren't actually "grid" like), and marginaleffects can also generate linear perditions at the mean.

Using the delta method is more computationally costly than using a linear combination (though marginaleffects is very efficient). Using linear combinations with orthogonal "grids" also often means that results from emmeans directly correspond to a models coefficients (which is a benefit for those who are used to looking at regressions tables to understand their models - this can be shown with an example).

...

Would you like me to add all of this in? Just some of this?

DominiqueMakowski: > Would you like me to add all of this in? Just some of this?

Anything you think is relevant, but I think it'll be good to be quite thorough and detailed here as the inner workings of marginal stuff are not often clearly explained so having details is good!

mattansb: Alright - I'll give it some time tomorrow.

mattansb:
I've added the expanded "technical details" section.

Some general comments:

**Statement of need**

I think the "statement of need" needs some restructuring, because the actual need it addresses is that emmeans and marginaleffects are sometimes hard to use - "the `modelbased` package is designed to be user-friendly, with a focus on simplicity and flexibility". That is, it doesn't provide any functionality in and of itself, but it provides a (very!) convenient wrapper with good defaults. 

**Predictions**

"For generalized linear models (GLMs), while the distinction between prediction and confidence intervals do not apply" - why not? I teach prediction intervals for non-gaussian models.
See also:
https://fromthebottomoftheheap.net/2017/05/01/glm-prediction-intervals-i/
https://fromthebottomoftheheap.net/2017/05/01/glm-prediction-intervals-ii/

**Marginal effects**

The different "marginals" need to be stated better here - the "marginal" in "marginal means" is not the same "marginal" that is in "marginal effects". 
In fact, I'm not sure, from reading the paper, if `estimate_contrasts()` gives marginal differences or average differences? Etc..

strengejacke:
> The different "marginals" need to be stated better here - the "marginal" in "marginal means" is not the same "marginal" that is in "marginal effects".
> In fact, I'm not sure, from reading the paper, if `estimate_contrasts()` gives marginal differences or average differences? Etc..

Since I updated my mixed models slides for teaching after working on _modelbased_, maybe we can use some of the "definitions" I use on my slides (hopefully, these are accurate):


After fitting a model, it is useful generate model-based estimates. Usually, there are three quantities of interest:

- *marginal means:* to better understand the relationship of predictors with the outcome. These are the average expected values, or adjusted predictions, of the response variable for different combinations of predictor values.
- *marginal effects:* to evaluate the average strength of an effect (slope, average coefficient).
- *marginal contrasts:* to look for (statistically significant) differences between  groups, which helps us, for instance, analyzing “social inequalities”.

The key difference:
- *marginal means* (or: adjusted predictions) return averages of the outcome variable, which allows you to say for instance “the average health score for a person at the age of sixty is 80 points”.
- *marginal effects* return averages of coefficients, which allows you to say, for instance, “the average effect of age (or ageing) on the health score is an decrease 5 points”.
- *marginal contrasts* return the average difference between two groups, typically indicated by different factor levels, which allows you to say “the average difference in health scores for lower and higher status groups at the same age is 15 points”.

Briefly:
- *marginal means* show the association of the dependent variable and one or more independent variables;
- *marginal effects* show you the average strength of a coefficient;
- *marginal contrasts* show you average differences in the outcome between groups

strengejacke:
> marginal differences or average differences? 

What would be the difference between these differences?

mattansb:
```R

library(marginaleffects)

mtcars$vs <- factor(mtcars$vs)
mtcars$gear <- factor(mtcars$gear)

mod <- glm(am ~ vs * hp * gear, 
           family = binomial(),
           data = mtcars)

# differences
comparisons(mod, variables = "vs",
            type = "response")

# average differences (mean unit level differences)
avg_comparisons(mod, variables = "vs",
                type = "response")
#> 
#>  Estimate Std. Error      z Pr(>|z|)   S  2.5 % 97.5 %
#>    -0.166      0.195 -0.851    0.395 1.3 -0.547  0.216
#> 
#> Term: vs
#> Type:  response 
#> Comparison: 1 - 0
#> 




# marginal means
avg_predictions(mod, variables = "vs",
                newdata = datagrid(hp = mean, gear = levels),
                type = "response")

# marginal differences (differences between marginal means)
avg_predictions(mod, variables = "vs",
                newdata = datagrid(hp = mean, gear = levels),
                type = "response",
                hypothesis = ~ reference)
#> 
#>  Hypothesis Estimate Std. Error     z Pr(>|z|)   S  2.5 % 97.5 %
#>       1 - 0   -0.302      0.239 -1.26    0.207 2.3 -0.772  0.167
#> 
#> Type:  response
#>  
```

mattansb:
I think we should avoid using the ambiguous term "marginal" as much as possible with regard to effects/slopes/differences (probably still okay with regard to "marginal means" since that can only be the "average" kind, not the "conditional" kind).

strengejacke:
Ok, I see, I think your distinction refers to the `marginalize` argument. I personally think, and maybe we should state that clearly, that we use "marginal" in terms of "marginalizing out" / "integrating out" / "averaging over/across". Then, marginal means/effects/contrasts makes sense. Whether the means/effects/contrasts refer to a specific person, an average person or the "broader population" (counterfactuals) depends on _how_ we "marginalizing out" / "integrating out" / "averaging over/across".

strengejacke:
Here are the _modelbased_ equivalents:

```r
library(modelbased)
library(marginaleffects)

mtcars$vs <- factor(mtcars$vs)
mtcars$gear <- factor(mtcars$gear)

mod <- glm(am ~ vs * hp * gear, family = binomial(), data = mtcars)

# average differences (mean unit level differences)
avg_comparisons(mod, variables = "vs")
estimate_contrasts(mod, "vs", marginalize = "population")

# marginal means over a regular grid
avg_predictions(mod, variables = "vs", newdata = datagrid(vs = levels, hp = mean, gear = levels))
estimate_means(mod, "vs", predict = "response")

# note the corrected prediction type
avg_predictions(mod, variables = "vs",
                newdata = datagrid(vs = levels, hp = mean, gear = levels),
                type = "invlink(link)")
estimate_means(mod, "vs")

# marginal differences (differences between marginal means)
avg_predictions(mod, variables = "vs",
                newdata = datagrid(hp = mean, gear = levels),
                type = "response",
                hypothesis = ~ reference)
estimate_contrasts(mod, "vs", comparison = ~ reference)
```

mattansb:
That's fine - but since there are these two meanings, I think it is best we stear clear of this term whenever possible. I would even try and find a better argument name for `marginalize` and explain in more detail what this argument does and why it's defaults are better / different than emmeans / marginaleffects'.

DominiqueMakowski:
> a better argument name for marginalize

any suggestions?

strengejacke:
> Doesn't margenalize control the "grid type"

Not fully. That's why Dom and I think it's clearer to avoid a rather "technical" approach to what's going on, and instead a more "practical", or: "what's the question I'd like to answer?" approach.

`marginalize` has two options (producing different data grids) that either produce results like `predict()` (predictions for a _specific_ subject, non-focals hold constant at mean and reference levels) or `emmeans()` (predictions for an _average_ subject, non-focals hold constant at mean or averaged over factor levels), and one option that does not produce a grid, but returns "counterfactual" predictions (predictions for a broader / more representative population, extrapolation to a hypothetical population).

In all three cases you get an estimated average value for your response, however, for different "persons"/"subjects"/"groups" - because in all three cases, non-focal terms are treated differently (according to how they're "averaged" - or "marginalized").

That's the idea behind this argument name, and to avoid a "technical" thinking of marginal means, and instead looking at predictions in terms of "what is the sample/entity/person we're making inferences about"?

paper/paper.md
```diff
@@ -30,57 +30,255 @@ affiliations:
...
 
+- Conceptual differences between predictions and marginal means?
+- Concept of focal and non-focal terms?
+- For marginalization, we can refer to non-focal terms, meaning it would be good to have the concept introduced somewhere
+
 ## Predictions
 
 At a fundamental level, `modelbased` and similar packages leverage model *predictions*.
+These predictions can be of different types, depending on the model and the question at hand.
+For instance, for linear regressions, predictions can be associated with **confidence intervals** (`predict="expectation"`) or **prediction intervals** (`predict="prediction"`).
```
strengejacke, line 60: Not only linear regression? Remove that subordinary sentence?

paper/paper.md
```diff
@@ -30,57 +30,255 @@ affiliations:
...
-In line with the `easystats`' *raison d'être*, the `modelbased` package is designed to be user-friendly, with a focus on simplicity and flexibility. 
+The two probably most popular R packages for extracting these quantities of interest from statistical models are `emmeans` (REF) and `marginaleffects` (REF). These packages are enormously feature rich and (almost) cover all imaginable needs for post-hoc analysis of statistical models. However, these packages are not always easy to use, especially for users who are not familiar with the underlying statistical concepts. The `modelbased` package aims to unlock this currently underused potential by providing a unified interface to extract marginal means, marginal effects, contrasts, comparisons and model predictions from a wide range of statistical models. It is built on top of the two aforementioned `emmeans` and `marginaleffects` packages. In line with the `easystats`' *raison d'être*, the `modelbased` package is designed to be user-friendly, with a focus on simplicity and flexibility.
 
 
 # Key concepts
```
strengejacke, line 50: We might add a short sentence clarifying the conceptual differences between predictions and marginal means?

paper/paper.md
```diff
@@ -30,57 +30,255 @@ affiliations:
...
+- `estimate_link()`: data grid, predictions on the link scale.
+
+*Note:* if the defaults are changed, then these functions can become redundant.
+For instance, `estimate_relation(..., predict="link")` will be equivalent to `estimate_link(...)`.
+
+These functions belong to the same family, and their relevance depends on the model and the research question.
+
 
-Mention types of predictions (expectations, predictions, response vs. the link scale)
+## Marginal means and effects
```
strengejacke, line 81: I think somewhere we can introduce the concept of focal and non-focal terms. For marginalization, we can refer to the wording of "non-focal terms", which is probably better than "non-focal predictors" or "other predictors", because sometimes, we have interaction _terms_.

DominiqueMakowski: Agreed

paper/paper.md
mattansb: Yes

mattansb:
Hmmm to me the word "marginalize" is _very_ technical, and `marginalize = "average"` is _very_ confusing. First, both options are potentially computing an average, depending on the non-focal predictor. Second, both words (can) mean "to average".
 
Since the options are "average over regular grid" or "average over the model frame" and the word "marginalize" has two meanings, I would have:
- `average = "regular"` / `average = "grid"` (what is currently `marginalize = "average"`)
- `average = "population"` (what is currently `marginalize = "population"`)

(Or to be _even more clear_, I would have `average_over=` as the argument name.)

strengejacke:
> Hmmm to me the word "marginalize" is _very_ technical, and `marginalize = "average"` is _very_ confusing. First, both options are potentially computing an average, depending on the non-focal predictor. Second, both words (can) mean "to average".
> 
> Since the options are "average over regular grid" or "average over the model frame" and the word "marginalize" has two meanings, I would have:
> 
> * `average = "regular"` / `average = "grid"` (what is currently `marginalize = "average"`)
> * `average = "population"` (what is currently `marginalize = "population"`)
> 
> (Or to be _even more clear_, I would have `average_over=` as the argument name.)

Sounds good, we should think about this - @DominiqueMakowski WDYT?

DominiqueMakowski:
There's two decisions to be made I think:

- Change name of **argument**: I'm somewhat positive about `average_over` even though it's a bit "out of scheme" and odd-looking. I prefer `average_over` much than `average` though. `average_over` / `marginalize` > `average`

- Change `"average"` option: I don't like `"grid"` because it's not indicative of what it's used for, as Daniel mentioned, I like the idea of focusing on the use-case rather than the technical thing that happens.  We toyed in the past with `"specific"`, but maybe we could do like `"average_individual"` / `"average_observation"` even though that could be too specific and inaccurate?

mattansb:
I'm confused by `marginalize = "average"` having anything to do with an "individual" - what individual? Only when none of the non-focal predictors are factors will the give an estimate of a single theoretical individual - else you are averaging over a grid of marginal means. 

What does `marginalize = "specific"` do? I haven't seen that documented.

strengejacke:
> What does marginalize = "specific" do? I haven't seen that documented.

That's `predict(newdata = <grid>)`. `marginalize="average"` is emmeans, and `marginalize="population"` is counterfactual (`avg_predictions(variables=...)`).

See also docs for `?estimate_means`:

Screenshot_20250130-204101

strengejacke:
> something that is weird to think about when averaging over a grid of marginal means

Why? How would you interpret the different types of averaging? I.e. to which subjects do your results refer to? (It's not helpful interpreting results in technical terms, I think, when a general audience should understand it)

mattansb:
Oh, now I get it - this isn't about averaging really, averaging is sometimes used, but it's about what units are the estimates generated (and then possibly averaged across).

So how about:

`marginalize = "specific"` -> `estimate_over = "individual"` (for a single unit)
`marginalize = "average"` -> `estimate_over = "average"` (for an average unit / average across marginal means)
`marginalize = "population"` -> `estimate_over = "population"` (for average across all units in a sample)

Or `estimate_for=` / `over=`

strengejacke:
> So how about:

Yes, that sounds feasible, too! 🙂

DominiqueMakowski:
We should also probably add in the paper & in the docstrings some examples about when people might want to use the non-default option for `estimate` (e.g., "population"). It's not obvious to me, and if we don't clarify that we might end up with an option that nobody ever uses just because they don't know they should

strengejacke:
We may add a vignette, I'd say, which is immediately online and independent from the package version on CRAN?

DominiqueMakowski:
Yes, though that is probably independent from throwing a few words in about potential usecases. Did you encounter scenarios where you preferred to use `estimate="population"` instead of "average"?

DominiqueMakowski:
interesting

DominiqueMakowski:
@IndrajeetPatil could you work your magick here and help us set the formatting / rendering of the paper? Forgot how to do that :p

strengejacke:
> interesting

It's useful for casual inference and (randomized) trials. Imagine you have 2 or more time points, and some drop outs over time. You could impute missing data, but that ideally requires MCAR. Often, drop out is rather systematic, like lower educated or less healthy persons drop out. Then you can compensate for this drop out either by calculating inverse probability weights or use g-computation to calculate accurate ATE/ATT/ATU.

rempsyc:
Please edit the `rmd` file from now on and not the `md` file

DominiqueMakowski:
Do we know why the references don't work?

image

paper/paper.Rmd
```diff
@@ -0,0 +1,305 @@
...
+)
+
+library(easystats)
+```
+
+# Statement of need
+
+Applied statistics have historically been focused on statistical *tests* (e.g., *t*-tests, correlation tests, ANOVAs), seen as most apt to provide researchers with interpretable answers to the questions they seek.
+However, these tests are typically based on statistical *models*, which are the true underlying cornerstone of modern data science.
+The replication crisis [@OSC2015estimating; @camerer2018evaluating] and methodological revolution(s) [@makowski2023we] have underlined the issues with the traditional focus on statistical tests (the effacement of model assumptions, a focus on null-hypothesis testing, non-compatibility with more complex variance structures), and the need to shift the focus to the models themselves.
```
strengejacke, line 95: Can we have one or two examples for "the issues"? Doesn't need to be full sentences.

paper/paper.Rmd
```diff
@@ -0,0 +1,308 @@
...
+
+*Note:* if the defaults are changed, then these functions can become redundant.
+For instance, `estimate_relation(..., predict = "link")` will be equivalent to `estimate_link(...)`.
+
+Which function to use depends on the model and the research question at hand (visualizing effects, generating predictions, etc.).
+
+
+## Marginal means and effects
+
+The concept of "marginal" in statistical modeling typically refers to the analysis of the effect of one or more "focus" variables while all other variables are held constant at specific values
```
strengejacke, line 142: I'm not quite happy with this first paragraph, because it focuses on "effects" (i.e. slopes), while we want to talk about relationship, comparisons/contrasts and effects/slopes. We should maybe focus on the "averaging" aspect (and maybe even move "contrasts" into an own section? - but that's not very necessary)

paper/paper.Rmd
```diff
@@ -0,0 +1,308 @@
...
+
+- `estimate_means()`: computes **Marginal Means**, i.e., the average predictions for each level of a categorical predictor, averaged across all levels of other predictors.
+- `estimate_contrasts()`: computes **Marginal Contrasts**, i.e., the comparison of marginal means at different levels of a factor to assess differences or effects.
+- `estimate_slopes()`: computes **Marginal Effects**, i.e., the change in the response variable for an infinitesimal change in a predictor, averaged across all levels of other predictors. They are essentially the partial derivatives of the response with respect to each predictor, useful for continuous predictors, and describe the average strength of an effect.
+
+
+**[TODO: details about types of marginalization]**
+
+
+## Technical details
```
strengejacke, line 156: Maybe we can shorten this section a bit? The paper becomes quite long, and I'm not sure if the technical details of the _backend_ packages are helpful to understand what **modelbased** is and how it works, and what's the idea behind our package?

paper/paper.Rmd
```diff
@@ -0,0 +1,308 @@
...
+
+The algorithmic heavy lifting is done by its two backend packages, `emmeans` and `marginaleffects` (the default),
+which can be set as a global option with (e.g., `options(modelbased_backend = "emmeans")`).
+
+Of the two, `emmeans` [@russell2024emmeans] is the more senior package and was originally known as `lsmeans` (for "Least-Squares Means").
+This term has been historically used to describe what are now more commonly referred to as "Estimated Marginal Means" or EMMs:
+predictions made over a regular grid—a counter-factual dataset containing all combinations of the categorical predictors in the model and typically the mean of numerical predictors.
```
strengejacke, line 163: I'm not convinced it's always a "counterfactual" data set ;-)

paper/paper.Rmd
```diff
@@ -0,0 +1,308 @@
...
+Note that `emmeans` can also use the delta method and can use non-regular grids,
+and `marginaleffects` can also generate linear predictions at the mean,
+but to obtain these requires some degree of competency in the relevant packages.
+
+Finally, `modelbased` leverages `get_datagrid()` function from the `insight` package [@ludecke2019insight]
+to intuitively generate an appropriate grid of data points for which predictions or effects or slopes will be estimated.
+And since these these packages support a wider range of models—including generalized linear models, mixed models, and Bayesian models, this means that `modelbased` also inherits the support for such models.
+
+# Examples
+
```
strengejacke, line 214: examples look good to me!

paper/paper.Rmd
DominiqueMakowski: lol you know best here

paper/paper.Rmd
DominiqueMakowski: yes we can streamline a bit, although in general I like the idea of having slightly more substance in that paper compared to some of our other ones (that whole domain deserves it and it will be useful as a reference to readers)

DominiqueMakowski:
Last thing is to copypaste the docstrings about `estimate=` and we're pretty much good to submit I think :)

rempsyc:
I made some changes to the writing. Some notes about the formatting:

1. Right now the output from `parameters::parameters(model)` is split over two lines, therefore repeating the first column twice. But not sure how to fix that unless to do it manually on the final md file before submitting.
2. The figure appears different somehow; a previous version plotted individual observations as well, but this version doesn't. What's the expectation here?
3. There's still this section to be completed: "TODO: details about types of marginalization"

mattansb:
> 1. Right now the output from parameters::parameters(model) is split over two lines, therefore repeating the first column twice. But not sure how to fix that unless to do it manually on the final md file before submitting.

That can be adjusted by setting `options(easystats_table_width = <numeric>)`.

strengejacke:
@DominiqueMakowski Feel free to revise the section I just worked on (wording, shorten/enhance). I think we can introduce the concepts as following:

- we have two types: predictions and marginal means (_Terminology_)
- we describe predictions (_Predictions_)
- we describe marginal means, pointing to the differences (_Marginal means_)
- we then say that we are _also_ able to look at strength of effect, not only their relationship (_Marginal effects_)
- if we want to know whether relationship of groups are statistically significantly different, we do this with contrasts (_Contrasts_, to do...)
- A special case is a third way of "averaging", `estimate = "population"`. This is in particular useful for counterfactual comparisons/contrasts (not sure how to name this section, to do...)

Then we have predictions/means, slopes, contrasts, marginalization, structured in a way so readers can follow it. WDYT?

rempsyc:
> That can be adjusted by setting `options(easystats_table_width = <numeric>)`.

Good point, but then it overflows the PDF (because the Parameter column is very large because of that value: `Petal Length × Species [versicolor]`). Maybe we just leave as is then.

strengejacke:
`print(..., select = "minimal")`

paper/paper.Rmd
strengejacke: Ok agreed

strengejacke:
@DominiqueMakowski A small section on _contrasts_ is still missing. Not sure how you prefer to frame the topic of predictions vs. marginal means, means/slopes/contrasts, and marginalization options? Feel free to completely revise that section ;-)

I thought it makes sense:
- to introduce into the difference between "predictions" and "marginal means"
- to say that if _associations_ are not the focus, but rather the strength of effect, we have _marginal effects_ or slopes
- we can compare groups/levels (_contrats_)
- interpretation of predictions is difficult, because there are different ways to calculate them, based on the same model and data (marginalization)
  - data grid based predictions
  - empirical distribution based predictions

Does this make sense? Take a look at the current state of the paper and feel free to change/move/modify to suit your idea of the paper.

strengejacke:
Ok, I already did some polishing on my part. If you're satisfied with the structure of the paper, I'd say there's not much missing, only final minor revisions - unless you want to re-write the Key-concepts section.

strengejacke:
@DominiqueMakowski I think you can merge this and open a new PR with a "cleaned" version of the paper.rmd that is ready for submission, and co-authors can have a final look before.

PR #360, opened 2025-01-20
Add proper control-info for `marginalize = "individual"`

strengejacke:


DominiqueMakowski:
Sorry for that late suggestion, but should we rename "individual" for "specific"? Because it's more that it's about "specific" observations (which in some contexts is indeed one "individual") than individuals 😬

DominiqueMakowski:
i.e., even though it's not a user-facing feature, "individual" might be misleading suggesting that it might have to do with marginalizing over participants or random factors or something like that

strengejacke:
Ok. Luckily, we had timing issues with the submission, so we may include this PR here, too, it also fixes an issue in footer printing for `estimate_relation()`.

strengejacke:
I noticed this because I'm currently updating my slides for teaching :-) This is probably a good stress test ;-)

PR #361, opened 2025-01-20
Prepare CRAN release, try to reduce time for examples

strengejacke:
@DominiqueMakowski I accidentally merged the initial submission PR, so I opened a new one. I tried to reduce running time for examples. Let's see if checks pass, and if so, you can submit from this PR.

PR #362, opened 2025-01-20
option to show/hide CI

strengejacke:
@DominiqueMakowski this is not urgent, but what do you think about having an option to remove the CI in plots?

### DIFF
--- a/R/visualisation_recipe_internal.R
+++ b/R/visualisation_recipe_internal.R
@@ -2,7 +2,7 @@
 
 
 #' @keywords internal
-.find_aes <- function(x) {
+.find_aes <- function(x, show_ci = TRUE) {
   data <- as.data.frame(x)
   data$.group <- 1
 
@@ -56,8 +56,10 @@
     # If x is a not-numeric, make pointrange
     if (is.numeric(data[[by[1]]])) {
       aes$type <- "ribbon"
-    } else {
+    } else if (isTRUE(show_ci)) {
       aes$type <- "pointrange"
+    } else {
+      aes$type <- "point"
     }
   }
   if (length(by) > 1) {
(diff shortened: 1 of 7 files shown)


R/visualisation_recipe_internal.R
```diff
@@ -237,6 +244,27 @@
     l <- l + 1
   }
 
+  # only points, no error bars - when show_ci = FALSE
```
DominiqueMakowski, line 247: I don't think you need a new separate layer when show_ci is FALSE, you can just replace the `pointrange` one's "geom" by `"point"`

Also I think currently if people modify  `point` list it modifies both the raw data points and these ones so it might be best to keep it under `pointrange`

strengejacke: Does it work to keep it under point range? I thought it might warn or error because of the ymin/ymax aes, which points don't have?

strengejacke: > Also I think currently if people modify `point` list it modifies both the raw data points and these ones so it might be best to keep it under `pointrange`

We could just skip modifying estimate points, but anyway, who wants to show row data and point geoms for predictions *without* error bars? ;-)
Hiding CIs is mostly intended for many line geoms, honestly.

DominiqueMakowski: > error because of the ymin/ymax aes, which points don't have?

no it doesn't 

```r
x <- modelbased::estimate_means(lm(Sepal.Length ~ Species, data = iris))
plot(x, pointrange = list(geom="point"))
```
image

Another option then is to allow for `ribbon = "none"` ?

strengejacke: That sounds good!  Let me try later

strengejacke: ok, I removed `show_ci` and just added a check for `ribbon = "none"` in #361

PR #366, opened 2025-01-26
Remove redundant label from group-variable

strengejacke:
Fixes #364
Fixes #368 

```r
library(modelbased)
data(efc, package = "modelbased")

# make categorical
efc <- datawizard::to_factor(efc, c("c161sex", "c172code", "e16sex"))
levels(efc$c172code) <- c("low", "mid", "high")
efc$e42dep <- as.factor(efc$e42dep)
fit <- lm(neg_c_7 ~ e16sex + c161sex * c172code + e42dep, data = efc)

estimate_contrasts(fit, c("c161sex", "c172code"), comparison = "pairwise")
estimate_contrasts(fit, c("c161sex", "c172code"), comparison = ratio ~ pairwise)

estimate_contrasts(fit, "c161sex", comparison = ~pairwise | c172code)
estimate_contrasts(fit, "c161sex", by = "c172code", comparison = ~pairwise)

estimate_contrasts(fit, "c161sex", comparison = ratio ~ pairwise | c172code)
estimate_contrasts(fit, "c161sex", by = "c172code", comparison = ratio ~ pairwise)

data(coffee_data, package = "modelbased")
m <- lm(alertness ~ time * coffee + sex, data = coffee_data)
estimate_contrasts(m, c("time", "coffee"), comparison = ratio ~ reference | coffee)
estimate_contrasts(m, "time", comparison = ratio ~ reference | coffee)
estimate_contrasts(m, "time", by = "coffee", comparison = ratio ~ reference)

estimate_contrasts(m, "time", comparison = ratio ~ sequential | coffee)
```

PR #367, opened 2025-01-27
Prepare CRAN release

strengejacke:


strengejacke:
@DominiqueMakowski I think this version now works for both the "old" and "new" marginaleffects version, yo you can try to submit again from this PR.

When the next marginaleffects version is out, we could probably simplify the formatting of contrasts (see #366), by simply putting the variables from `by` into the formula, e.g. `~pairwise | <by>`. Then filtering is automatically done and we only have the relevant comparisons.

This makes things a lot easier, **but**: at the moment, this does not filter specific levels for `by` when we do `by = "gear='5'"`. We can try to find a solution for this, then we could format _any_ comparison supported by _marginaleffects_, but for now, we should submit this version here to CRAN.

PR #372, opened 2025-01-30
`estimate_contrasts()` for `estimate_relation()` etc

strengejacke:
When we want contrasts for random effects, _marginaleffects_ returns too small standard errors (at least for *glmmTMB*), while `estimate_relation()` relies on the SEs from `predict()` (which are the best we can get, according to the package authors).

Thus, we implement a very basic contrast method, which, however works for random effects, and thereby allows MAIHDA analysis with the _modelbased_ package.


DominiqueMakowski:
MAIDHA stands for?

strengejacke:
https://www.sciencedirect.com/science/article/pii/S235282732400065X

Latest shit in epidemiology

strengejacke:
See also https://strengejacke.github.io/ggeffects/articles/practical_intersectionality.html

strengejacke:
> See also https://strengejacke.github.io/ggeffects/articles/practical_intersectionality.html

You can now fully replace _ggeffects_ with _modelbased_ in that vignette. 🥂

PR #386, opened 2025-02-09
Add `residualize_over_grid()`

strengejacke:


bwiernik:
Could we instead make this a `model_residuals()` function that takes a `newdata` argument, which could be a grid from get_datagrid()?

strengejacke:
Sure, but do we need `newdata`? The grid is extracted / taken from the `estimate_means()` (or related) result.

DominiqueMakowski:
It can still be useful for power-users to have a way to pass the data no?

strengejacke:
@gemini-code-assist review

strengejacke:
This is the current state - though I'm not sure how to fix the last error, and I don't understand why ggplot complains about discrete values?

strengejacke:
``` r
library(modelbased)

model <- lm(Sepal.Length ~ Species * Sepal.Width, data = iris)

pr <- estimate_means(model, c("Sepal.Width", "Species"))
plot(pr, show_residuals = TRUE)
```



``` r


set.seed(1234)
x1 <- rnorm(200)
x2 <- rnorm(200)

# quadratic relationship
y <- 2 * x1 + x1^2 + 4 * x2 + rnorm(200)
d <- data.frame(x1, x2, y)

model <- lm(y ~ x1 + x2, data = d)

pr <- estimate_means(model, "x1")
plot(pr, show_residuals = TRUE)
```



``` r

model <- lm(y ~ x1 + x2, data = d)

pr <- estimate_means(model, c("x1", "x2=[sd]"))
plot(pr, show_residuals = TRUE)
#> Error in `scale_colour_continuous()`:
#> ! Discrete value supplied to a continuous scale.
#> ℹ Example values: -0.93, 0.07, and 1.08.
```

<sup>Created on 2025-08-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
> It can still be useful for power-users to have a way to pass the data no?

I'm not quite sure at which point we should / can add `newdata`, since the data frame method already accepts a grid? Feel free to modify/add this function, where necessary.

PR #387, opened 2025-02-10
Plot 2nd numeric in by as continuous or factor, based on number of unique values

### LINKED ISSUE #201: Plot method for interactions
First, there is an issue in `get_datagrid()` (https://github.com/easystats/insight/issues/611), but once this is fixed, the order of columns should be as shown in the below example. For such plots, I would expect the first `at` predictor at the x-axis, the second as group-factor, i.e. I would expect what _ggeffects_ returns in this case. I don't think the current shown plot behaviour from _modelbased_ is correct.

``` r
m <- lm(Sepal.Width ~ Petal.Length + Species * Petal.Width, data = iris)
grid <- insight::get_datagrid(m, at = "Species * Petal.Width", range = "grid", preserve_range = FALSE)
head(grid)
#>   Petal.Width    Species Petal.Length
#> 1   0.4370957     setosa        3.758
#> 2   1.1993333     setosa        3.758
#> 3   1.9615710     setosa        3.758
#> 4   0.4370957 versicolor        3.758
#> 5   1.1993333 versicolor        3.758
#> 6   1.9615710 versicolor        3.758

modelbased::estimate_expectation(m, data = grid[c(2,1,3)]) |> plot()
#> geom_path: Each group consists of only one observation. Do you need to adjust
#> the group aesthetic?
```

``` r
ggeffects::ggpredict(m, c("Species", "Petal.Width")) |> plot()
```

<sup>Created on 2022-08-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

strengejacke:
Fixes #201 

```r
library(modelbased)

set.seed(1234)
x1 <- rnorm(200)
x2 <- rnorm(200)
# quadratic relationship
y <- 2 * x1 + x1^2 + 4 * x2 + rnorm(200)

d <- data.frame(x1, x2, y)
model <- lm(y ~ x1 + x2, data = d)

pr <- estimate_means(model, c("x1", "x2"))
plot(pr)
```

image

```r
pr <- estimate_means(model, c("x1", "x2 = [sd]"))
plot(pr)
```

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: modelbased
 Title: Estimation of Model-Based Predictions, Contrasts and Means
-Version: 0.9.0.8
+Version: 0.9.0.9
 Authors@R:
     c(person(given = "Dominique",
              family = "Makowski",
@@ -92,4 +92,4 @@ Roxygen: list(markdown = TRUE)
 Config/Needs/check: stan-dev/cmdstanr
 Config/Needs/website: easystats/easystatstemplate
 LazyData: true
-Remotes: easystats/insight
+Remotes: easystats/insight, easystats/see
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,9 @@
 * The deprecated function `visualisation_matrix()` has been removed. Use
   `insight::get_datagrid()` instead.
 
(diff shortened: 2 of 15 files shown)


strengejacke:
I think this makes more sense to have a "discrete" legend if explicitly only few values are requested.

PR #406, opened 2025-02-19
Draft case study on causal inference

strengejacke:


DominiqueMakowski:
looks interesting!

strengejacke:
_modelbased_ is listed in the Causal Inference task view on CRAN, so it maybe makes sense to make this topic a bit more prominent - we're currently working on a proposal where we plan to apply these methods, I think it becomes more common in epi-research, where you often have observational data, but still some kind of "intervention" or "treatment" characteristic that interests you.

btw, we should ask Ben Bolker if he could add _modelbased_ to the mixed models task view, too.

PR #443, opened 2025-03-11
estimate_grouplevle(): return dispersion for Bayesian models

DominiqueMakowski:


strengejacke:
Have you run the full test? It is skipped for the latest CRAN version of insight and only works with the very latest dev version

DominiqueMakowski:
Why do I keep posting errors without updating to latest first 🤦 


This however falls on latest:

``` r
library(rstanarm)
#> Loading required package: Rcpp
#> This is rstanarm version 2.32.1
#> - See https://mc-stan.org/rstanarm/articles/priors for changes to default priors!
#> - Default priors may change, so it's safest to specify priors, even if equivalent to the defaults.
#> - For execution on a local, multicore CPU with excess RAM we recommend calling
#>   options(mc.cores = parallel::detectCores())

d <- iris
d$Group <- as.factor(rep(c("G1", "G2", "G3"), each = 50))
# m <- lme4::lmer(Sepal.Width ~ Petal.Width + (Petal.Width | Group), data = d)
# m <- glmmTMB::glmmTMB(
#   Sepal.Width ~ Petal.Width + (Petal.Width | Group),
#   data = d
# )
m <- rstanarm::stan_lmer(
  Sepal.Width ~ Petal.Width + (Petal.Width | Group),
  data = d,
  refresh = 0
)
#> Warning: There were 5 divergent transitions after warmup. See
#> https://mc-stan.org/misc/warnings.html#divergent-transitions-after-warmup
#> to find out why this is a problem and how to eliminate them.
#> Warning: Examine the pairs() plot to diagnose sampling problems
modelbased::estimate_grouplevel(m)
#> Error in `$<-.data.frame`(`*tmp*`, "Level", value = c("(Intercept)", "Petal.Width", : replacement has 12 rows, data has 11
```

<sup>Created on 2025-03-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Do we need to explicitly add these arguments? We could document these like for other functions. See `...` in `estimate_means()`: https://easystats.github.io/modelbased/reference/estimate_means.html

DominiqueMakowski:
> Do we need to explicitly add these arguments? 

It's just that they are not the defaults of parameters (which don't return dispersion but do return tests etc.)

DominiqueMakowski:
Merge?

DominiqueMakowski:
Mmh I think before cleaning-up we need to fix rstanarm processing parameters:

1. Specifying `effects="random"` errors
2. We don't pull the Intercept-related random effects
3. (The "Group" inside `attributes(params)$clean_parameters`doesn't seem to be well inferred)

``` r
m <- insight::download_model("stanreg_lmerMod_5")

lme4::ranef(m)
#> $Species
#>            (Intercept)       Cat1Y
#> setosa      -0.9586980  0.05587158
#> versicolor   0.1213769  0.04254976
#> virginica    0.8723191 -0.09720587
#> 
#> with conditional variances for "Species"
parameters::parameters(m, effects = "random", group_level=TRUE)
#> Error in attr(tables[[type]], "digits") <- digits: attempt to set an attribute on NULL
parameters::parameters(m, effects = "all", group_level=TRUE)
#> # Fixed Effects
#> 
#> Parameter   | Median |        95% CI |     pd |  Rhat |     ESS |              Prior
#> ------------------------------------------------------------------------------------
#> (Intercept) |   1.20 | [ 0.30, 2.05] | 99.20% | 1.000 | 1257.00 | Normal (0 +- 7.62)
#> Cat1Y       |  -0.03 | [-0.43, 0.33] | 63.18% | 1.002 | 1295.00 | Normal (0 +- 1.91)
#> 
#> # Random Effects: Species
#> 
#> Parameter           | Median |         95% CI |     pd |  Rhat |     ESS
#> ------------------------------------------------------------------------
#> Species:setosa      |  -0.96 | [-1.82, -0.06] | 98.02% | 1.000 | 1247.00
#> Species:setosa      |   0.06 | [-0.32,  0.44] | 70.10% | 1.002 | 1372.00
#> Species:versicolor  |   0.12 | [-0.73,  1.03] | 62.52% | 1.000 | 1255.00
#> Species:versicolor  |   0.04 | [-0.31,  0.45] | 67.22% | 1.001 | 1315.00
#> Species:virginica   |   0.87 | [ 0.03,  1.78] | 97.67% | 1.000 | 1258.00
#> Species:virginica   |  -0.10 | [-0.49,  0.28] | 81.15% | 1.002 | 1309.00
#> (Intercept)         |   0.45 | [ 0.16,  1.61] |   100% | 1.001 | 1936.00
#> Cat1Y ~ (Intercept) |  -0.03 | [-0.28,  0.19] | 72.82% | 1.001 | 2204.00
#> Cat1Y               |   0.03 | [ 0.00,  0.50] |   100% | 1.003 | 1385.00
#> 
#> Uncertainty intervals (equal-tailed) computed using a MCMC distribution
#>   approximation.
```

<sup>Created on 2025-03-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
It's maybe also an issue form brms.

strengejacke:
Seems like `group_level` only works for `effects = "all"`, that's why we do it like that.

strengejacke:
ok, let me check tests, and then we can merge.

PR #446, opened 2025-03-12
Strengejacke/issue444

### LINKED ISSUE #444: `estimate_grouplevel()` bug
Why do I keep posting errors without updating to latest first 🤦 

This however falls on latest:

``` r
library(rstanarm)
#> Loading required package: Rcpp
#> This is rstanarm version 2.32.1
#> - See https://mc-stan.org/rstanarm/articles/priors for changes to default priors!
#> - Default priors may change, so it's safest to specify priors, even if equivalent to the defaults.
#> - For execution on a local, multicore CPU with excess RAM we recommend calling
#>   options(mc.cores = parallel::detectCores())

d <- iris
d$Group <- as.factor(rep(c("G1", "G2", "G3"), each = 50))
# m <- lme4::lmer(Sepal.Width ~ Petal.Width + (Petal.Width | Group), data = d)
# m <- glmmTMB::glmmTMB(
#   Sepal.Width ~ Petal.Width + (Petal.Width | Group),
#   data = d
# )
m <- rstanarm::stan_lmer(
  Sepal.Width ~ Petal.Width + (Petal.Width | Group),
  data = d,
  refresh = 0
)
#> Warning: There were 5 divergent transitions after warmup. See
#> https://mc-stan.org/misc/warnings.html#divergent-transitions-after-warmup
#> to find out why this is a problem and how to eliminate them.
#> Warning: Examine the pairs() plot to diagnose sampling problems
modelbased::estimate_grouplevel(m)
#> Error in `$<-.data.frame`(`*tmp*`, "Level", value = c("(Intercept)", "Petal.Width", : replacement has 12 rows, data has 11
```

<sup>Created on 2025-03-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

_Originally posted by @DominiqueMakowski in https://github.com/easystats/modelbased/issues/443#issuecomment-2715819733_

strengejacke:
Fixes #444

### DIFF
--- a/R/estimate_grouplevel.R
+++ b/R/estimate_grouplevel.R
@@ -69,12 +69,20 @@ estimate_grouplevel <- function(model, type = "random", dispersion = TRUE, test 
     ...
   )
 
-  # Re-add info
-  if (!"Group" %in% names(params)) {
-    params$Group <- attributes(params)$clean_parameters$Group
+  # get cleaned parameter names with additional information
+  clean_parameters <- attributes(params)$clean_parameters$Group
+
+  # fix for rstanarm, which contains a sigma columns
+  if (inherits(model, "stanreg")) {
+    clean_parameters <- clean_parameters[clean_parameters$Component != "sigma", ]
   }
-  if (!"Level" %in% names(params)) {
-    params$Level <- attributes(params)$clean_parameters$Cleaned_Parameter
+
+  # Re-add info
+  if (!"Group" %in% names(params) && !is.null(clean_parameters)) {
+    params$Group <- clean_parameters$Group
+  }
+  if (!"Level" %in% names(params) && !is.null(clean_parameters)) {
+    params$Level <- clean_parameters$Cleaned_Parameter
   }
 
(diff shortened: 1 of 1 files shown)


R/estimate_grouplevel.R
```diff
@@ -69,12 +69,20 @@ estimate_grouplevel <- function(model, type = "random", dispersion = TRUE, test
     ...
   )
 
+  # get cleaned parameter names with additional information
+  clean_parameters <- attributes(params)$clean_parameters$Group
+
+  # fix for rstanarm, which contains a sigma columns
+  if (inherits(model, "stanreg")) {
+    clean_parameters <- clean_parameters[clean_parameters$Component != "sigma", ]
```
DominiqueMakowski, line 77: In theory there can be models where you have random effects on `sigma` à-la-brms, but I quickly checked rstanarm's functions and I don't think there is currently any model implemented that would allow that.

However, this might change in the future so we might want to make a note of it or think of a more robust solution

PR #464, opened 2025-04-04
Simplify reshape_grouplevel()

DominiqueMakowski:


DominiqueMakowski:
Yes, previous docs used the term "list" colloquially, but it should be a character vector indeed

and yes it's easier to keep group to be one thing (although technically we could still allow for multiple groups and have a long format but I think in most practical applications people are interested in reshaping one group at a time (often to merge with original data)

strengejacke:
Following tests are failing, likely related to this PR:

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-estimate_grouplevel.R:18:3'): estimate_grouplevel - lme4 ─────
nrow(reshape_grouplevel(random)) (`actual`) not equal to nrow(data) (`expected`).

  `actual`:  18
`expected`: 180
── Failure ('test-estimate_grouplevel.R:24:3'): estimate_grouplevel - lme4 ─────
nrow(reshape_grouplevel(random)) (`actual`) not equal to nrow(mtcars) (`expected`).

  `actual`:  6
`expected`: 32
── Failure ('test-estimate_grouplevel.R:30:3'): estimate_grouplevel - lme4 ─────
nrow(reshape_grouplevel(random)) (`actual`) not equal to nrow(data) (`expected`).

  `actual`:  18
`expected`: 180
── Failure ('test-estimate_grouplevel.R:45:3'): estimate_grouplevel - lme4 ─────
nrow(reshaped) (`actual`) not equal to nrow(data) (`expected`).

  `actual`:  18
`expected`: 180
```

strengejacke:
I'm not sure, but I think the major difference is that `merge()` is missing. The current `reshape_grouplevel()` also only returns a much shorter table (less nrows than data), but "merges" the wide-reshaped (shorter) table into the original data:

```r
data <- merge(data, data_wide, by = g, sort = FALSE)
```

Is that right?

R/reshape_grouplevel.R
```diff
@@ -1,80 +1,76 @@
...
+    ))
+    group <- group[1]
+  }
+  x <- x[x$Group %in% group, ]
 
-    # Merge while preserving order of original random
-    data <- attributes(x)$data # Get original dataframe of random
-    data[["__sort_id"]] <- seq_len(nrow(data))
-    data <- merge(data, data_wide, by = g, sort = FALSE)
-    data <- data[order(data[["__sort_id"]]), ]
```
strengejacke, line 69: I think we still need something like this `merge()` in the code, to match the wide table, which is shorter than the original data, to the original data.

DominiqueMakowski: I don't think we need to match to the original data now, otherwise it kind of disregards the `group` that was selected 

In other words, the resulting dataframe should match the number of individual items in the requested `group` (and we can leave it to the user to merge it to the original dataframe if need be)

DominiqueMakowski:
> I'm not sure, but I think the major difference is that merge() is missing. 

Yes that is the goal, this was actually an unwanted behavior. Say you have a repeated measure dataframe with 100 participants that each did 200 trials. Your model is like `~ ... + (1|Participant) + (1|Item)`. 

You are interested in extracting the Median random factor for the participants with `reshape_grouplevel(group="Participant")`. With the merge, you would always return a dataframe of 100 * 200 rows, which is a lot of unnecessary duplication and makes it unwieldy. Now, the return is, as would one expect, a dataframe with 100 rows (one for each participant), and the user can then merge it back easily to the original one if need be

strengejacke:
ok, can you then update failing tests? I think you can better validate the expected output/results

DominiqueMakowski:
Should we "clean" up (Intercept) to be `Intercept`? I wonder if that wasn't the case before 

``` r
model <- lme4::lmer(mpg ~ wt + (1 | gear) + (1 | carb), data = mtcars)
modelbased::estimate_grouplevel(model)
#> Group | Level | Parameter   | Coefficient |   SE |        95% CI
#> ----------------------------------------------------------------
#> carb  | 1     | (Intercept) |        0.55 | 0.68 | [-0.77, 1.87]
#> carb  | 2     | (Intercept) |        0.29 | 0.64 | [-0.96, 1.54]
#> carb  | 3     | (Intercept) |       -0.02 | 0.73 | [-1.46, 1.42]
#> carb  | 4     | (Intercept) |       -0.50 | 0.64 | [-1.76, 0.75]
#> carb  | 6     | (Intercept) |       -0.12 | 0.78 | [-1.65, 1.40]
#> carb  | 8     | (Intercept) |       -0.20 | 0.78 | [-1.72, 1.33]
#> gear  | 3     | (Intercept) |       -0.54 | 0.71 | [-1.93, 0.84]
#> gear  | 4     | (Intercept) |        1.22 | 0.77 | [-0.28, 2.73]
#> gear  | 5     | (Intercept) |       -0.68 | 0.95 | [-2.55, 1.19]
modelbased::reshape_grouplevel(modelbased::estimate_grouplevel(model), group = "gear")
#>   gear CI_high_(Intercept) CI_low_(Intercept) Coefficient_(Intercept)
#> 1    3           0.8444319         -1.9311665              -0.5433673
#> 2    4           2.7250501         -0.2766538               1.2241982
#> 3    5           1.1850160         -2.5466776              -0.6808308
#>   SE_(Intercept)
#> 1      0.7080738
#> 2      0.7657549
#> 3      0.9519801
```

<sup>Created on 2025-04-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
I think for brms, it's named `Intercept`. But we can rename, for consistency

PR #474, opened 2025-04-10
[Revision]

### LINKED ISSUE #458: JOSS Review 1
- [x] Your text kind of half-implies that emmeans does NOT use the Delta method by default. This would be an incorrect reading. Both packages use the Delta method. The difference is that marginaleffects uses numeric differentiation to get there, and that it also supports bootstrap and simulation-based inference.
- [x] I strongly disagree with the argument that marginaleffects sacrifices simplicity for power: For example, the most important quantity that any of our packages computes is arguably the Average Treatment Effect: `avg_comparisons(model)` That's not hard or complicated! I'm not saying that modelbased is more complicated, just that the difference in syntax is a purely aesthetic matter of preference and convention, rather than an ordering of complexity.

DominiqueMakowski:
Fixes #458


strengejacke (approved):

DominiqueMakowski:
Also tagging @vincentarelbundock in case there's any further suggestions :)

strengejacke:
How can we include changes into the paper submitted to JOSS?

DominiqueMakowski:
I think at the end it will auto regenerate it no? Otherwise there's probably a `generate` command for their editorial bot

strengejacke:
Ah, they grab it from this repo? Can't quite remember, it's been a while since my last JOSS submission ;-)

DominiqueMakowski:
> Ah, they grab it from this repo?

yup

PR #490, opened 2025-04-25
JOSS reviewer 2

DominiqueMakowski:
> I would add that estimation (via models) has been offered as an alternative/complement to hypothesis tests, and I would cite The New Statistics manifesto https://pubmed.ncbi.nlm.nih.gov/24220629/ as a touchstone here.

- [x] done

> But why would I learn modelbased instead of just working even harder at learning marginaleffects? I think the other selling points for modelbased worth mentioning is 1) a unified syntax for estimation that can switch between backends, 2) helpers for plotting the estimates and 3) consistency with the rest of the easystats 'verse.

- [x] done

> "[29] But the models themselves withhold even more usefulness" - withhold means something more like hide or conceal and I don't think models are hiding extra usefulness.

- [x] done

> "[39] The modelbased package, built on top of these two packages, aims to unleash this vast, untapped potential by providing a unified interface to extract marginal means, marginal effects," - "unleash this vast, untapped potential" is a little over the top.

- [x] done

> except for one glaring sentence ("[81] can be considered as “marginal” pendant to estimate_relation()")

- [x] todo 

> and one spot that should have a citation ("shrinkage prevents overfitting and [151] improves generalizability."). I appreciate it as a code vignette. However, much of this will have to be cut for length.

- [x] todo


DominiqueMakowski:
We can remove the "backend" details section and say that we removed this lengthy section, and hopefully it will be okay ?

IndrajeetPatil:
You can also get LLMs to opine on whether we can make some prose more succinct.

strengejacke:
> We can remove the "backend" details section and say that we removed this lengthy section, and hopefully it will be okay ?

You mean the section `Technical Details`?

strengejacke:
I think we can remove it and make a vignette out of it. But I leave the final decision to you.

strengejacke:
(I added the vignette, it's good to have it, independent from whether we keep it in the paper or not...)

PR #542, opened 2025-07-26
inequality ratio can be pairwise

strengejacke:
- [x] Update vignette with an example?
- [x] add table with formula notation to vignette
- [x] add tests
- [x] add examples to docs?
- [x] check if it's also that simple to add pairwise comparison to slopes


strengejacke:
@gemini-code-assist review

PR #544, opened 2025-07-29
`by` can be of length two for inequality-contrasts

strengejacke:


strengejacke:
@gemini-code-assist review

PR #545, opened 2025-07-31
Message when numeric is treated as integer

strengejacke:


strengejacke:
@gemini-code-assist review

strengejacke:
> I think describing it as treating as "continuous" vs "numeric" would be clearer

What do you exactly mean? I'm not sure which part of the message to modify.

Here's the original message:

"Numeric variable appears to be ordinal or Likert-scale (integer values, no more than 5 unique values) and is treated as discrete variable. Set `integer_as_numeric = TRUE` to disable this check and always treat numeric variables as continuous."

bwiernik:
Added some specific places. I don't think that "numeric" implies "continuous numeric" clearly and so we should use "treat as continuous" as the description of what is happening

PR #558, opened 2025-08-27
automatic detection of `by` variables now also works with in-formula transformations

strengejacke:


strengejacke:
@gemini-code-assist review

PR #638, opened 2026-07-01
emmeans backend doesn't seem to allow stratification by random factors

DominiqueMakowski:
``` r
library(modelbased)

dat <- lme4::sleepstudy
dat$Condition <- ifelse(dat$Days < median(dat$Days), "Low", "High")
m <- lme4::lmer(Reaction ~ Condition + (Condition | Subject), data = dat)

get_marginalcontrasts(m, contrast = "Condition", by = "Subject")
#>         Parameter Subject Difference       SE     CI_low    CI_high           t
#> 1  (Low) - (High)     308 -96.163212 8.620562 -113.17754 -79.148883 -11.1550981
#> 2  (Low) - (High)     309  -9.114864 8.620562  -26.12919   7.899465  -1.0573400
#> 3  (Low) - (High)     310 -20.485587 8.620562  -37.49992  -3.471258  -2.3763633
...


get_emcontrasts(m, contrast = "Condition", by = "Subject")
#> Error:
#> ! Variable `Subject` only used as random effects grouping factor and
#>   cannot be used with `backend = "emmeans"`. Please use `backend =
#>   "marginaleffects"` instead.
```

<sup>Created on 2026-07-01 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

So instead we throw a more informative error.

Unless @mattansb you are aware of some hidden backdoor to emmeans to get that (ai says *recover_data() deliberately keeps only the fixed-effects predictors in the reference grid. Grouping variables used solely for random effects are discarded.*).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: modelbased
 Title: Estimation of Model-Based Predictions, Contrasts and Means
-Version: 0.16.0.1
+Version: 0.16.0.2
 Authors@R:
     c(person(given = "Dominique",
              family = "Makowski",
--- a/NEWS.md
+++ b/NEWS.md
@@ -29,6 +29,13 @@
 
 * Minor bug fix when evaluating `slope` and `by` arguments in `estimate_slopes()`.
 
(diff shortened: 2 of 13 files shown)


mattansb:
That is correct - emmeans currently only uses the fixed effects for constructing the reference grid.

PR #662, opened 2026-09-11
Move check-functions into own R file, refactor longer checks into separate functions

strengejacke:


strengejacke:
All local checks pass

